PR #22, opened 2024-02-22
Implement lobster source filters kind and prefix

paulboege:
- implement lobster source filters kind and prefix
- add integration test at integration-tests/projects/filter
- add source filter description to docs/config_files.md

### DIFF
--- a/docs/config_files.md
+++ b/docs/config_files.md
@@ -34,6 +34,14 @@ The `source` attribute assigns a LOBSTER file to contribute to this
 level. There is some extra information you can supply with the `with`
 keyword.
 
+Using `with kind <string>` allows to filter items with `kind` in the 
+LOBSTER source file. For TRLC requirements the kind attribute in the 
+LOBSTER source file corresponds to the TRLC requirements type.
+
+Using `with prefix <string>` allows to filter items with a certain
+prefix in the `tag` attribute in the LOBSTER source file. For TRLC 
+requirements the tag attribute corresponds to the requiremet name.
+
 Specifically for requirements you can say:
 
 ```
@@ -42,8 +50,6 @@ requirements "Requirements" {
      valid_status {"Valid", "Approved"};
 }
 ```
-
-We expect to add other `with` properties in the future.
 
 #### trace to
 
--- a/lobster/io.py
+++ b/lobster/io.py
@@ -110,14 +110,24 @@ def lobster_read(mh, filename, level, items, source_info=None):
         else:
             item = Activity.from_json(level, raw, data["version"])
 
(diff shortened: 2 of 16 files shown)


lobster/io.py
```diff
@@ -110,14 +110,24 @@ def lobster_read(mh, filename, level, items, source_info=None):
...
         if source_info is not None:
             item.perform_source_checks(source_info)
-
-        items[item.tag.key()] = item
+    
+            # evaluate source_info filters
+            for f, v in source_info['filters']:
+                if f == 'prefix':
+                    filter_conditions.append(item.tag.tag.startswith(v))
+                if f == 'kind':
```
phiwuu, line 122: I know this has been done all over the place, but do you see a chance to introduce constants, such that the string `kind` does not need to be created twice? It is like a hard-coded number (although it's not a number), and it is needed in `parser.py`as well. As I said, this coding style has been done very often so far, but I would like to improve our style and get rid of magic strings.

phiwuu (commented):
Neat feature!

lobster/io.py
paulboege: For introducing string constants I don't see a straight forward approach:
- An Item is parsed in the from_json() method after reading the LOBSTER / json file. 
- The config parser code is independent of the Item and classes derived from Item. However the config syntax contains attributes from Item. This already existed with the 'with valid_status'. With adding 'with prefix' and 'with kind' this got worse.

To improve the filter implementation I will move the code you cite above from io.py to items.py, Item.perform_source_checks() and return a boolean. Then the definition of filter conditions are part of the Item class.

Do you agree?

phiwuu: I see your point. Don't worry, keep it as it is. Otherwise we would mix a feature implementation with a refactoring, which should be separate PRs with separate reviews, and especially with separate commits in the history.

docs/config_files.md
```diff
@@ -34,6 +34,14 @@ The `source` attribute assigns a LOBSTER file to contribute to this
...
 keyword.
 
+Using `with kind <string>` allows to filter items with `kind` in the 
+LOBSTER source file. For TRLC requirements the kind attribute in the 
+LOBSTER source file corresponds to the TRLC requirements type.
+
+Using `with prefix <string>` allows to filter items with a certain
+prefix in the `tag` attribute in the LOBSTER source file. For TRLC 
+requirements the tag attribute corresponds to the requiremet name.
+
```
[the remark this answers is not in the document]
paulboege: done

phiwuu (approved):

phiwuu (approved):

Issue #26, opened 2024-02-29
Improve visual appearance of HTML report with respect to accessibility


phiwuu:
Feedback from users with read-green weakness is that, they cannot distinguish the background coloring of the lobster HTML report. For some red and gray look very similar, for others it is red and green.

1. use different colors
2. use icons additionally to indicate "error", "justified exemption" and "ok".

PR #34, opened 2024-04-22
improve accessibility in LOBSTER html report

ankushsonare1:
improve accessibility in LOBSTER report: HTML report #26

As a maintainer of the LOBSTER tool

I change prominent colours.

So that visually impaired people (color blindness) can distinguish easily between error and success


mugdhadhole1:
Colors seems ok
Ok to merge

florianschanda:
I suggest to add a command-line flag to do this, or even a javascript enabled toggle in the document itself.

Changing the default colours for everyone is not intuitive.

florianschanda:
I suggest to add a command-line flag to do this, or even a javascript enabled toggle in the document itself.

Changing the default colours for everyone is not intuitive.

christophkloeffel (changes requested):
please remove report.lobster and lobster_report.html

report.lobster
```diff
@@ -0,0 +1,13353 @@
+{
```
phiwuu, line 1: Please don't add this file. I guess it is a mistake?

lobster_report.html
```diff
@@ -0,0 +1,11218 @@
+<!DOCTYPE html>
```
phiwuu, line 1: Please don't add this file. I guess it is a mistake?

phiwuu (changes requested):

phiwuu:
The new images look very nice.

phiwuu:
Please clarify the copyright of the new icons. If we are OK to use them, then please update the colours as follows:

1. use a darker green instead of light green, e.g. `b2e1b2`
2. change red to purple, e.g. `ffb2ff`
3. keep the gray as is

florianschanda (dismissed):
I suggest to add a command-line flag to do this, or even a javascript enabled toggle in the document itself.

Changing the default colours for everyone is not intuitive.

phiwuu:
I saw your comment, but I don't get the point. What do you mean with "not intuitive"?
The color changes are minimal, but solve the issue for visually impaired people. Adding a command-line flag or javascript is much more effort compared to changing the colors. We want to close this issue and move on.

ankushsonare1:
I have check new svg icons its not copyright.

florianschanda:
Traffic light colours are universally clear.
The proposed colours are not - you need to add a legend at the very least.

florianschanda:
I would still strongly argue for a command-line option if a live, in document option using JS is too hard to do. It won't be too hard to do given how we bake in the style sheet anyway.

lobster/tools/core/html_report.py
```diff
@@ -286,13 +286,13 @@ def write_html(fd, report, dot):
         "padding-left"  : "0.2em",
     }
     doc.style[".item-ok"] = {
-        "background-color" : "#efe",
+        "background-color" : "#b2e1b2",
```
phiwuu, line 289: please add a switch case here, as suggested by @florianschanda

lobster/tools/core/html_report.py
```diff
@@ -286,13 +286,13 @@ def write_html(fd, report, dot):
...
     }
     doc.style[".item-partial"] = {
         "background-color" : "#ffe",
     }
     doc.style[".item-missing"] = {
-        "background-color" : "#fee",
+        "background-color" : "#ffb2ff",
```
phiwuu, line 295: and here as well

phiwuu (changes requested):

lobster/tools/core/html_report.py
```diff
@@ -465,6 +465,9 @@ def main():
                     help="path to dot utility (https://graphviz.org), \
                     by default expected in PATH",
                     default=None)
+    ap.add_argument("--color_blind",
```
christophkloeffel, line 468: please use flag `--color-blind`

lobster/tools/core/html_report.py
```diff
@@ -465,6 +465,9 @@ def main():
...
+                        action="store_true",
+                        help="Enable color blind mode")
```
christophkloeffel, line 470: be more specific in the help message: 
something like "uses a color palette with a higher contrast"

christophkloeffel (changes requested):

christophkloeffel:
please also create an entry in the changelog

christophkloeffel (approved):

phiwuu (approved):

Issue #36, opened 2024-04-30
lobster-python: false positive on duplicate function definition


jnow-87:
lobster-python reports a duplicated function definition for `severity()` in the following example:
```
import dataclasses

@dataclasses.dataclass
class Diagnostic: 
    _severity: int = dataclasses.field()

    @property
    def severity(self):
        return self._severity

    @severity.setter
    def severity(self, value: int):
        self._severity = value
```

These are clearly not duplicate functions but separate ones which can both be called on the same instance if Diagnostic. Hence, this case should be fixed within lobster.

The problem might originate from here https://github.com/bmw-software-engineering/lobster/blob/6bb274e19b47b8ced6fc9c6d103c5e4418cf83a7/lobster/tools/python/python.py#L197-L198 since the function tag is built only using `self.fqn()`, which is probably the function name. A solution might be to consider the entire function signature as the tag.

phiwuu:
Possible duplicate of https://github.com/bmw-software-engineering/lobster/issues/30

Issue #30, opened 2024-03-28
Handling duplicate function definitions


phiwuu:
### Current Behaviour
If a Python file contains several function definitions with the same name, then the output of `lobster-python` cannot be forwarded to `lobster-report`.

Consider the following Python file (call it `code.py`):
```python
value = False

if value:
    def compute():
        return 42
else:
    def compute():
        return 43

print(compute())
```
Now run `lobster-python`:
```bash
> lobster-python code.py --out="python.lobster"
```
The result of `python.lobster` is the following:
```json
{
  "data": [
    {
      "tag": "python code.compute",
      "location": {
        "kind": "file",
        "file": "code.py",
        "line": 4,
        "column": null
      },
      "name": "code.compute",
      "messages": [],
      "just_up": [],
      "just_down": [],
      "just_global": [],
      "language": "Python",
      "kind": "Function"
    },
    {
      "tag": "python code.compute",
      "location": {
        "kind": "file",
        "file": "code.py",
        "line": 8,
        "column": null
      },
      "name": "code.compute",
      "messages": [],
      "just_up": [],
      "just_down": [],
      "just_global": [],
      "language": "Python",
      "kind": "Function"
    }
  ],
  "generator": "lobster_python",
  "schema": "lobster-imp-trace",
  "version": 3
}
```
Now `lobster-report` cannot handle this input, as it detects a duplicate definition of "code.compute".
The error is raised here: https://github.com/bmw-software-engineering/lobster/blob/666f8afed28a7e4a5156fc7cacb47f2b6998c622/lobster/io.py#L128

### Expected Behaviour
Now the question is: What is the expected behaviour?
For sure the above Python code is valid and needs to be supported by lobster.
On the other hand, duplicate definitions shall be detected, too.
We could add a unique identifier to the function name, like the line number, but then it will be hard to reference this function in other elements of the tracing policy. But the V-model of ISO 26262 always recommends to trace from code to requirements, so I wonder if tracing from something to python code through lobster-links is a real use case. Then adding line numbers might be a valid solution. This could be implemented as a switch flag.

Any ideas how to handle this best?

phiwuu:
Real-world example: https://github.com/python-attrs/cattrs/blob/v22.1.0/src/cattrs/_compat.py#L184

christophkloeffel:
we could just add the line_nr at the end of the current Tracing Tag to make it unique.
[...]

PR #39, opened 2024-06-14
adds line number to tracing tag of python functions to make them unique

christophkloeffel:
fixes:
* bmw-software-engineering/lobster#36
* bmw-software-engineering/lobster#30

### DIFF
--- a/lobster/tools/python/python.py
+++ b/lobster/tools/python/python.py
@@ -90,7 +90,11 @@ class Python_Traceable_Node:
             rv = self.parent.fqn() + "."
         else:
             rv = ""
-        rv += self.name
+        if self.location.line is not None and \
+          isinstance(self, Python_Function):
+            rv += f"{self.name}:{str(self.location.line)}"
+        else:
+            rv += self.name
         return rv
 
     def lobster_tag(self):


phiwuu (approved):

lobster/tools/python/python.py
```diff
@@ -90,7 +90,11 @@ def fqn(self):
             rv = self.parent.fqn() + "."
         else:
             rv = ""
-        rv += self.name
+        if self.location.line is not None and \
```
phiwuu, line 93: Please consider the following refactoring:

1. Call `self.fqn` once before creating `Implementation`, see https://github.com/bmw-software-engineering/lobster/blob/86cd606a0d624e663c57fa298b88a11f5f2ed4a8/lobster/tools/python/python.py#L157
2. Find a unique name by iterating over `items`
3. Don't use the line number, because this ID will be dynamic and changes if someone adds new lines to the file. Instead, use some kind of counter.

The first tag name shall always be without a number. Only the second tag name shall add a suffix.

phiwuu (changes requested):

phiwuu (approved):

Issue #31, opened 2024-04-12
Support using graphviz with explicit path


phiwuu:
### Problem:
When using Lobster with [Bazel](https://bazel.build/), it clears the `PATH` variable (to make sure all inputs are exactly known),
But `lobster-html-report` calls graphviz by looking up the location of the graphviz executible in `PATH`. Hence it is not possible to generate an image of the tracing policy in the HTML report.

### Request:
Add a feature to specify the path to the graphviz executable.

PR #41, opened 2024-06-14
adds --dot argument to specify the dot executable path

christophkloeffel:
implements: #31

### DIFF
--- a/lobster/tools/core/html_report.py
+++ b/lobster/tools/core/html_report.py
@@ -36,9 +36,9 @@ from lobster.items import (Tracing_Status, Item,
 LOBSTER_GH = "https://github.com/bmw-software-engineering/lobster"
 
 
-def is_dot_available():
+def is_dot_available(dot):
     try:
-        subprocess.run(["dot", "-V"],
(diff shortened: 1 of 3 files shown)


lobster/tools/core/html_report.py
```diff
@@ -461,6 +461,9 @@ def main():
                     default="report.lobster")
     ap.add_argument("--out",
                     default="lobster_report.html")
+    ap.add_argument("--dot",
+                    help="path to dot utility, by default expected in PATH",
```
phiwuu, line 465: just a minor comment: not everybody knows what "dot" is, and these users might be confused what this help message means. A google search for "dot" does not immediately point to `https://graphviz.org/`. How about the following message:
`help="path to dot utility (https://graphviz.org/), by default expected in PATH",`

phiwuu (approved):
Pull request approved, but see recommendation about "help" message.

phiwuu (approved):

lobster/tools/core/html_report.py
christophkloeffel: done

phiwuu (approved):

Issue #60, opened 2024-07-31
lobster-online-report tool shall work with submodules


TannazVhdBMWExt:
`lobster-online-report` does work fine with main modules but if the file has been chosen from a submodule the git repository URL is not correct.

PR #63, opened 2024-07-31
lobster-online-report works with sub/ mainmodules

TannazVhdBMWExt:
the git-meta-information has been
adapted

Resolves: https://github.com/bmw-software-engineering/lobster/issues/60

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 
 ### 0.9.17
 
+* The `lobster-online-report` tool now works with config files located in 
+  main- and submodules of a repository
+
 * The `lobster-python` tool now adds the line number to the function
   identifier. This supports situations where different functions have
   the same name
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ### 0.9.18-dev
 
+* The `lobster-online-report` tool now works with config files located in 
+  main- and submodules of a repository. This feature needs `git 1.7.8` or higher.
+
 * The `lobster-codebeamer` tool now uses codebeamer api v3.
   Please note that the api v3 returns the value "Unset" for a codebeamer
   item status if the status is actually empty. The api v1 did not return
@@ -11,14 +14,7 @@
   resulting lobster file will now contain "Unset" as status information,
   too, instead of `Null`.
 
-### 0.9.18-dev
-
-
-
 ### 0.9.17
-
-* The `lobster-online-report` tool now works with config files located in 
-  main- and submodules of a repository
 
 * The `lobster-python` tool now adds the line number to the function
   identifier. This supports situations where different functions have
--- a/lobster/tools/core/online_report.py
+++ b/lobster/tools/core/online_report.py
@@ -36,36 +36,86 @@ def is_git_main_module(path):
     return os.path.isdir(os.path.join(path, ".git"))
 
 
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 
 ### 0.9.17
 
+* The `lobster-online-report` tool now works with config files located in
```
phiwuu, line 8: Same here, we are at `0.9.18-dev` already

phiwuu (changes requested):

lobster/tools/core/online_report.py
```diff
@@ -32,6 +32,48 @@ class Parse_Error(Exception):
     pass
 
 
+def is_git_main_module(path):
+    return os.path.isdir(os.path.join(path, ".git"))
+
+
+def find_repo_main_root(file_path):
```
phiwuu, line 39: I did a bit of research. It seems the most robust way is to do the following:
1. first check if the folder is inside a git submodule using `--show-superproject-working-tree`
2. if not, then check if it is inside a git working tree using `--show-toplevel`

This is code that our AI generated:
```python
import os
import subprocess

def is_dir_in_git_submodule(directory):
    """
    Checks if a given directory is nested inside a Git submodule.
    
    Args:
        directory (str): The path to the directory to check.
    
    Returns:
        bool: True if the directory is inside a Git submodule, False otherwise.
    """
    try:
        # Check if the directory is part of a Git submodule
        result = subprocess.run(['git', 'rev-parse', '--show-superproject-working-tree'], cwd=directory, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
        if result.returncode == 0 and result.stdout.strip():
            return True
        else:
            return False
    except (subprocess.CalledProcessError, OSError):
        return False

def is_dir_in_git_main_module(directory):
    """
    Checks if a given directory is nested inside a Git main module.
    
    Args:
        directory (str): The path to the directory to check.
    
    Returns:
        bool: True if the directory is inside a Git main module, False otherwise.
    """
    try:
        # Check if the directory is part of a Git main module
        result = subprocess.run(['git', 'rev-parse', '--show-toplevel'], cwd=directory, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
        if result.returncode == 0 and result.stdout.strip():
            return True
        else:
            return False
    except (subprocess.CalledProcessError, OSError):
        return False
```

phiwuu (changes requested):

lobster/tools/core/online_report.py
```diff
@@ -32,6 +32,98 @@ class Parse_Error(Exception):
...
+        result = subprocess.run(['git', 'rev-parse', '--show-toplevel'],
+                                cwd=directory,
+                                stdout=subprocess.PIPE,
+                                stderr=subprocess.PIPE,
+                                universal_newlines=True,
+                                check=True)
+        if result.returncode == 0 and result.stdout.strip():
+            return True, result.stdout.strip()
+        else:
+            return False, ''
```
phiwuu, line 92: You could also return `None` instead of `''`.
But actually one return value would be enough. Instead of 
```python
return True, result.stdout.strip()
```
just do
```python
return result.stdout.strip()
```
and instead of
```python
return False, ''
```
just do
```python
return None
```

phiwuu (approved):

Issue #45, opened 2024-06-26
`lobster-codebeamer` shall support upstream references


phiwuu:
There is no way to verify the traces from codebeamer requirements to other codebeamer requirements, because the output file of `lobster-codebeamer` does not contain a `refs` section.

Below is an excerpt of an example `codebamer.lobster` file that now contains a `ref` section:
```json
  "data": [
    {
      "tag": "req 649259",
      "location": {
        ...
      },
      "name": "example requirement",
      "messages": [],
      "just_up": [],
      "just_down": [],
      "just_global": [],
      "framework": "codebeamer",
      "kind": "Technical Requirement",
      "text": null,
      "status": "Valid"
      "refs": [
        "req 12345",
        "req 67890"
      ],
    },
```

This requires some kind of configuration. In the above example the codebeamer item `649259` can have an attribute called `derived from`, and contains links to items `12345` and `67890`.

Furthermore, there is no way to specify the following values:
- just_up
- just_down
- just_global

I have created https://github.com/bmw-software-engineering/lobster/issues/50 for these configuration parameters. The current ticket shall only focus on the `refs`  section.

phiwuu:
Note: We should implement https://github.com/bmw-software-engineering/lobster/issues/49 first.

PR #68, opened 2024-08-07
lobster-codebeamer supports refs upstream reference

TannazVhdBMWExt:
An optional config file as argument has been added to lobster-codebeamer. This config file should includes a list of upstream references and respective cb-fieldnames

Resolves https://github.com/bmw-software-engineering/lobster/issues/45


lobster/tools/codebeamer/codebeamer.py
```diff
@@ -193,6 +215,27 @@ def to_lobster(cb_config, cb_item):
...
+                    flat_values_list = list(value for custom_field in cb_item["customFields"]
+                                            if custom_field["name"] == displayed_name and custom_field.get("values")
+                                            for value in custom_field["values"])
+                if not flat_values_list:
+                    continue
+
+                map_reference_name_to_function[reference_name](req, flat_values_list)
+
+    if "name" not in cb_item:
+        req.error("Item lacks a summary text")
```
phiwuu, line 237: This would re-introduce the limitation to allow only items with a summary. Please remove it.

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -210,17 +253,50 @@ def import_tagged(mh, cb_config, items_to_import):
...
+        provided_config_keys = set(data.keys())
+        supported_references = set(SUPPORTED_REFERENCES)
+
+        if not provided_config_keys.issubset(supported_references):
+            raise Exception("The provided references are not supported! "
+                            "supported referenes: '%s'" % ', '.join(SUPPORTED_REFERENCES))
+
+        json_config = {}
+        for key, value in data.items():
+            json_config[key] = ensure_array_of_strings(value)
```
phiwuu, line 279: Why is the call to `ensure_array_of_strings` needed? Under which circumstances is `value` a list, and when is it not a list? And why does it have to be a list?

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -210,17 +253,50 @@ def import_tagged(mh, cb_config, items_to_import):
...
+
+
+def parse_cb_config_file(file_name):
+    assert isinstance(file_name, str)
+    assert os.path.isfile(file_name)
+
+    with open(file_name, "r") as file:
+        data = json.loads(file.read())
+
+        provided_config_keys = set(data.keys())
```
phiwuu, line 270: You can unindent everything after `json.loads(file.read())`. The file can be closed immediately once reading the file has finished.

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -210,17 +253,50 @@ def import_tagged(mh, cb_config, items_to_import):
...
     modes = ap.add_mutually_exclusive_group(required=True)
     modes.add_argument("--import-tagged",
                        metavar="LOBSTER_FILE",
                        default=None)
+
     modes.add_argument("--import-query",
                        metavar="CB_QUERY_ID",
                        default=None)
 
+    ap.add_argument("--config-file",
```
phiwuu, line 295: Please name it just `--config` to keep it short, and to keep it aligned with all the other lobster tools.

phiwuu (changes requested):
Please add the following piece of information to `/packages/lobster-tool-codebeamer/README.md`.
- a description of the `--config-file` parameter
- an example of a possible configuration

lobster/tools/codebeamer/codebeamer.py
TannazVhdBMWExt: Sure

Just for information: 

- "trlc" also uses "--config-file" as an argument
- "report" uses "--lobster-config" as an argument

lobster/tools/codebeamer/codebeamer.py
TannazVhdBMWExt: If the user uses more than one fieldname for one reference.
Right now our code works with these three variant of config files:

{
"refs" : **["derived from", "derived from2", "derived from3", "derived from4"]**
}
 or
{
"refs" : **"derived from"**
}
 or
{
"refs" : **["derived from"]**
}

TannazVhdBMWExt: I used 'ensure_array_of_strings' to ensure that the field names are always a list, because I want to work with **_one_** data type in the rest of the code.

phiwuu (approved):

PR #70, opened 2024-08-21
`lobster-codebeamer` uses authentication token

TannazVhdBMWExt:


phiwuu (commented):
OK, but merge conflicts need to be resolved.

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -280,10 +302,9 @@ def parse_cb_config(file_name):
                         "supported referenes: '%s'" %
                         ', '.join(SUPPORTED_REFERENCES))
 
-    json_config = {}
     for key, value in data.items():
         json_config[key] = ensure_array_of_strings(value)
```
phiwuu, line 306: `json_config["references"][key] = ensure_array_of_strings(value)`

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -332,14 +354,18 @@ def main():
...
         "timeout"    : options.timeout,
     }
 
     if options.config:
         if os.path.isfile(options.config):
-            cb_config["references"] = parse_cb_config(options.config)
+            cb_config["references"], config_token = (
+                parse_cb_config(options.config))
+            if not cb_config["token"] and config_token:
+                cb_config["token"] = config_token
```
phiwuu, line 368: Simplify like this:
```python
if os.path.isfile(options.config):
    cb_config.update(parse_cb_config(options.config))
else:
    ap.error("cannot open config file '%s'" % options.config)
```
This implies that the config file always has precedence over the command line arguments.
That might be counter-intuitive. We can also do it the other way around:
```python
    cb_config = {}
    if options.config:
        if os.path.isfile(options.config):
            cb_config =parse_cb_config(options.config)
        else:
            ap.error("cannot open config file '%s'" % options.config)

    cb_config.update({
        "root"       : options.cb_root,
        "base"       : "%s/cb/api/v3" % options.cb_root,
        "user"       : options.cb_user,
        "pass"       : options.cb_pass,
        "token"      : options.cb_token,
        "verify_ssl" : not options.ignore_ssl_errors,
        "page_size"  : options.query_size,
        "timeout"    : options.timeout,
    })
```

phiwuu (changes requested):

phiwuu (changes requested):
Please also update the part where the `.netrc` is read:
```python
if (not cb_config[TOKEN]) and (cb_config["user"] is None or cb_config["pass"] is None):
 ...
```

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -50,6 +50,8 @@
 from lobster.errors import Message_Handler, LOBSTER_Error
```
phiwuu, line 50: Currently the token can ONLY be provided in the configuration file. But in `README.md` you write that it can be provided as a command line argument, too.

phiwuu (approved):

phiwuu (approved):

PR #85, opened 2024-09-30
Add integration tests to ci

DiFerMa:


phiwuu (commented):
I believe the `MAKEFILE` needs an update since the introducion of `lobster-cpptest`.
Please have a look at https://github.com/bmw-software-engineering/lobster/blob/dda6fda8464fda00e9b2ae4fede9cde76674bd9c/Makefile#L25-L33 and add the following line:
```
make -C packages/lobster-cpptest
```

.github/workflows/ci.yml
```diff
@@ -72,3 +71,20 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install python version
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - name: Install dependencies
+        run: |
+          python3 -m pip install --upgrade pip --break-system-packages
+          python3 -m pip install -r requirements_dev.txt --break-system-packages
```
christophkloeffel, line 87:
```suggestion
          python3 -m pip install --upgrade pip
          python3 -m pip install -r requirements_dev.txt
```

christophkloeffel: since we are installing python explicitly via GitHub action there is no need for this argument to be present

christophkloeffel (changes requested):

christophkloeffel:
CI fails because of #89

christophkloeffel:
will be replaced by #96

Issue #10, opened 2023-08-24
allow optional up-linking


florianschanda:
It shall be possible to define a tracing policy where several upstream targets can be specified via `trace to:`, but the coverage shall be satisfied as soon as one link to at least one target is found.

Example:
```
activity "test" {
  trace to: "system requirement" or "software requirement";
}
```

Trace to itself shall be allowed.

phiwuu:
Note: The parsing of the "trace to" keyword is happening here: https://github.com/bmw-software-engineering/lobster/blob/c87654f6fe5dca0e6155f2a3912495f99f840076/lobster/config/parser.py#L192

Currently the config parser uses the `needs_tracing_down` attribute to remember if a level needs traces downwards. This is  parsed here: https://github.com/bmw-software-engineering/lobster/blob/c87654f6fe5dca0e6155f2a3912495f99f840076/lobster/config/parser.py#L188
A possible refactoring is to use a list of levels instead of a Boolean value. The list will then include all levels.

**But it is unclear how to compute the "downstream" coverage if one of the incoming connections is optional.**

phiwuu:
# Summary of use cases
Here is a summary of use cases that we need to support.

## Multiple Database Alternatives
Imagine a project allows several database alternatives for their software requirements, like TRLC, Codebeamer, IBM Doors.
The tracing policy will look like this:
```mermaid
flowchart BT
    PR[Product Requirements]
    Solution1[Codebeamer]
    Solution2[TRLC]
    Solution3[IBM Doors]
    SWR[Implementation]
    SWR-->|Path A|Solution1-->|Path A|PR
    SWR-->|Path B|Solution2-->|Path B|PR
    SWR-->|Path C|Solution3-->|Path C|PR
```
All three paths A, B, and C must be possible individually. That means, one of the paths could be used, but is not mandatory.

phiwuu:
New proposal:

The configuration file shall only support the following rules:

1. `trace to` specifies all the outgoing targets.
2. `trace from` specifies all the incoming targets.
3. Multiple lines of the same rule are treated as if they had an `and` between them.
4. `or` is supported in each line, to specify that only one target is needed, not all of them.
5. Self-references are also allowed.

Advantages:

- This means, the logic to implement for `trace to` and `trace from` is the same, except that the direction is inversed.
- The `trace to` keyword is backwards compatible.

Disadvantage:
It is a breaking change, because `requires` will no longer be supported, and `trace from` will be added.

PR #91, opened 2024-10-09
Support "or" in uplinks

TannazVhdBMWExt:
https://github.com/bmw-software-engineering/lobster/issues/10

### DIFF
--- a/tests-integration/projects/basic/report.reference_output
+++ b/tests-integration/projects/basic/report.reference_output
@@ -555,7 +555,7 @@
     "System Requirements": {
       "name": "System Requirements",
       "kind": "requirements",
-      "traces": [],
+      "trace_to": [],
       "source": [
         {
           "file": "system-requirements.lobster",
@@ -565,9 +565,7 @@
           ]
         }
       ],
-      "needs_tracing_up": false,
-      "needs_tracing_down": true,
-      "breakdown_requirements": [
+      "trace_from": [
         [
           "Software Requirements"
         ]
@@ -576,7 +574,7 @@
     "Software Requirements": {
       "name": "Software Requirements",
       "kind": "requirements",
-      "traces": [
+      "trace_to": [
         "System Requirements"
       ],
       "source": [
@@ -586,9 +584,7 @@
           "valid_status": []
         }
       ],
-      "needs_tracing_up": true,
-      "needs_tracing_down": true,
-      "breakdown_requirements": [
+      "trace_from": [
         [
           "Code"
         ],
@@ -600,7 +596,7 @@
     "Code": {
       "name": "Code",
       "kind": "implementation",
-      "traces": [
+      "trace_to": [
         "Software Requirements"
       ],
       "source": [
@@ -612,15 +608,13 @@
           "file": "pycode.lobster",
           "filters": []
         }
-      ],
-      "needs_tracing_up": true,
-      "needs_tracing_down": false,
-      "breakdown_requirements": []
+      ]
+      "trace_from": []
     },
     "Verification Test": {
       "name": "Verification Test",
       "kind": "activity",
-      "traces": [
+      "trace_to": [
         "Software Requirements"
       ],
       "source": [
@@ -637,9 +631,7 @@
           "filters": []
         }
       ],
-      "needs_tracing_up": true,
-      "needs_tracing_down": false,
-      "breakdown_requirements": []
+      "trace_from": []
     }
   },
   "matrix": []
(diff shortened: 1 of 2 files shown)


phiwuu (changes requested):
- please update the appropriate `README.md`
- please update the `CHANGELOG.md`
- please update https://github.com/bmw-software-engineering/lobster/blob/main/documentation/config_files.md

mugdhadhole1 (changes requested):
When running lobster-ci-report with the attached inputs, it exits with
 return code 0 where it should exit with return code 1.

In attached example, there are 2 functions in python and 2 requirements in different trlc files, the tracing policy says that `trace to: "Requirements1";`
But one function traces to "Requirements2" so, the expectation is that lobster-ci-report treats this as an error and returns 1, but in fact it returns 0.
The reason is that, lobster-ci-report only checks if there are error messages attached to any items it does not evaluate the status, in this example the status is "PARTIAL" but the item has no message. 
So, possible solutions

- lobster-ci-report does not only check the error messages but the status as well.
- lobster-report attaches a message to the item. (preferred solution)

[Example.zip](https://github.com/user-attachments/files/17714548/Example.zip)

phiwuu:
@TannazVhdBMWExt, this pull request shows no checks at all. We need to investigate the reason. Why are CI jobs not running?

phiwuu:
@TannazVhdBMWExt please rebase on main and ensure that checks are running.

lobster/config/parser.py
```diff
@@ -102,13 +102,11 @@ def parse_level_declaration(self):
                        "duplicate declaration")
 
         item = {
-            "name"                   : level_name,
-            "kind"                   : level_kind,
-            "traces"                 : [],
```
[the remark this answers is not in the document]
phiwuu: Note: if "dot" is not installed, then the bug will not be visible, because the tracing policy graph will not be rendered. See https://github.com/bmw-software-engineering/lobster/blob/f69c82ba408c7282fe48988b0a98ef106dee9816/lobster/tools/core/html_report.py#L365-L369

lobster/items.py
```diff
@@ -168,9 +179,7 @@ def determine_status(self, config, stab):
                 self.tracing_status = Tracing_Status.JUSTIFIED
             else:
                 self.tracing_status = Tracing_Status.OK
-        elif (ok_up or ok_down) and \
-             level["needs_tracing_up"] and \
-             level["needs_tracing_down"]:
+        elif ok_up or ok_down:
             self.tracing_status = Tracing_Status.PARTIAL
```
mugdhadhole1, line 183: The Tracing_Status should be "OK" only if no constraint was given in the tracing policy, but with the current implementation the status is "PARTIAL"

packages/lobster-core/README.md
```diff
@@ -14,11 +14,41 @@ You can generate a report linking everything together with `lobster-report`.
 The report is in JSON, but you can generate more readable versions of it
 with additional tools:
```
mugdhadhole1, line 15: lobster-report is not an additional tool to itself, hence these lines are inconsistent with the below line

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Changelog
 
+### 1.0.0-dev
+
+* The `lobster-report` tool now supports optional "OR" in uplinks (`trace to`). In 
+  the `lobster.config` file the keyword `requires` has been changed to `trace from`.
```
phiwuu, line 8:
```suggestion
* The behavior of `trace to` and `requires` in the configuration of `lobster-report` has been changed.
  `trace to` no longer adds an implicit condition in the reverse direction.
  All trace conditions have to be specified explicitly.
  The reverse direction can be specified with `trace from`.
  The keyword `requires` is no longer supported.
```

phiwuu (changes requested):

lobster/tools/core/html_report/html_report.py
```diff
@@ -135,10 +135,11 @@ def create_policy_diagram(doc, report, dot):
 
     for level in report.config.values():
         source = name_hash(level["name"])
-        for target in map(name_hash, level["traces"]):
+        all_trace_tos = sum(level["trace_to"], [])
+        for target in map(name_hash, all_trace_tos):
             # Not a mistake; we want to show the tracing down, whereas
             # in the config file we indicate how we trace up.
-            graph += '  n_%s -> n_%s;\n' % (target, source)
+            graph += '  n_%s -> n_%s;\n' % (source, target)
```
kedarnn, line 142: Any specific reason why the source and target are interchanged?

PR #93, opened 2024-10-10
Packaging correction of `lobster-cpptest` tool

DiFerMa:


christophkloeffel:
CI fails because the linter found errors in the files in submodule "parser" of the `lobster-cpptest` tool. Please fix those linter issues first.

Make sure you always add an `__init__.py` file, otherwise `make lint` will not include that module in the check. That's why these errors got merged on the main branch.

phiwuu (approved):

PR #97, opened 2024-10-13
uses build instead of invoking setup.py directly

christophkloeffel:
https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html


phiwuu (approved):
Ok, but what is the advantage of this change?

Issue #102, opened 2024-10-17
Catch `UnicodeDecodeError` in `lobster-json`


phiwuu:
Opening input files, which are not encoded with UTF-8, might lead to an uncaught `UnicodeDecodeError` error.
This is not convenient for the user, only software developers will understand it.

I recommend to catch the error, print a meaningful message, and exit with error code `1`.

This is the location where the error is raised:
https://github.com/bmw-software-engineering/lobster/blob/948471a180487f564f12fce8e7b9737f7d9822fa/lobster/tools/json/json.py#L111

See also https://github.com/bmw-software-engineering/lobster/issues/100

PR #103, opened 2024-10-17
Catch `UnicodeDecodeError` in `lobster-json`

phiwuu:
The `UnicodeDecodeError` is catched, and an error message is printed.
Afterwards `lobster-json` exits.

Issue: #102

### DIFF
--- a/lobster/tools/json/json.py
+++ b/lobster/tools/json/json.py
@@ -53,14 +53,11 @@ def get_item(root, path, required):
         elif required:
             raise Malformed_Input("object does not contain %s" % field,
                                   root)
-        else:
-            return None
+        return None
 
     elif required:
         raise Malformed_Input("not an object", root)
-
-    else:
-        return None
+    return None
 
 
 def syn_test_name(file_name):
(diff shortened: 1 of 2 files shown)


lobster/tools/json/json.py
```diff
@@ -107,15 +104,15 @@ def process_tool_options(self, options, work_list):
...
-        # First we follow the test-list items to get the actual data
-        # we're interested in.
         try:
+            with open(file_name, "r", encoding="UTF-8") as fd:
+                data = json.load(fd)
             data = get_item(root     = data,
                             path     = options.test_list,
                             required = True)
+        except UnicodeDecodeError as decode_error:
+            print("%s: File is not encoded in utf-8: %s" % (file_name, decode_error))
```
kedarnn, line 114: I observed that the old string formatting method  is used rather than f-strings, is there any specific reason due to which it is preferred?

phiwuu: No, there is no particular reason. I just used it because it is used everywhere else, to stay consistent.

phiwuu: This is something we should address in the coding guideline https://github.com/bmw-software-engineering/lobster/pull/88

christophkloeffel (approved):

christophkloeffel (approved):

PR #113, opened 2024-11-04
Add make tracing for TRCL tool

DiFerMa:


.github/workflows/docs.yml
```diff
@@ -37,13 +37,12 @@ jobs:
           python-version: "3.9"
       - name: Install dependencies
         run: |
+          sudo apt-get install -y graphviz
           python -m pip install --upgrade pip
           pip install -r requirements_dev.txt
```
christophkloeffel, line 42:
```suggestion
          pip install -r requirements_dev.txt
          echo "$GITHUB_WORKSPACE" >> $GITHUB_PATH
```

add the workspace folder to PATH

Makefile
```diff
@@ -90,3 +90,37 @@ test-ci: system-tests unit-tests coverage
 docs:
 	rm -rf docs
 	mkdir docs
+	@-make tracing
+
+tracing: report.lobster
+	mkdir -p docs
+	make lobster/html/assets.py
+	./lobster-html-report report.lobster --out=docs/tracing.html
```
christophkloeffel, line 98:
```suggestion
	lobster-html-report report.lobster --out=docs/tracing.html
```
remove "./" on all occurrences. adding the workspace folder to PATH solves this globally

.github/workflows/docs.yml
```diff
@@ -37,13 +37,12 @@ jobs:
...
       - name: Generate docs
         run: |
           make docs
-        env:
-          PYTHONPATH: "."
```
christophkloeffel, line 46: add this in order to avoid problems in the future and for a consistent structure, also used in TRLC at this step in the CI

christophkloeffel (changes requested):

christophkloeffel (approved):

PR #114, opened 2024-11-04
Refactoring for lobster/report.py file. Modifications in html_report …

kedarnn:
…as per the refactoring changes

### DIFF
--- a/lobster/report.py
+++ b/lobster/report.py
@@ -22,8 +22,6 @@ import json
 from collections import OrderedDict
 from dataclasses import dataclass
 
-# from dataclasses import dataclass
-
 from lobster.items import Tracing_Status, Requirement, Implementation, Activity
 from lobster.config.parser import load as load_config
 from lobster.errors import Message_Handler
@@ -33,10 +31,10 @@ from lobster.location import File_Reference
 
 @dataclass
 class Coverage:
-    level: str
-    items: int
-    ok: int
-    coverage: None
+    level    : str
+    items    : int
+    ok       : int
+    coverage : None
 
 
 class Report:
@@ -176,6 +174,16 @@ class Report:
         self.compute_items_and_coverage_for_items(data)
 
     def compute_items_and_coverage_for_items(self, data):
+        """
+        Function calcuates items and coverage for the items
+        Parameters
+        ----------
+        data - contents of lobster json file.
+
+        Returns - None
+        -------
+
+        """
         self.config = data["policy"]
         for level in data["levels"]:
             assert level["name"] in self.config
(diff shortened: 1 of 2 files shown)


lobster/report.py
```diff
@@ -20,6 +20,9 @@
 import os.path
 import json
 from collections import OrderedDict
+from dataclasses import dataclass
+
+# from dataclasses import dataclass
```
mugdhadhole1, line 25: Is this comment necessary? If not we should remove it

lobster/report.py
```diff
@@ -28,6 +31,14 @@
 from lobster.location import File_Reference
 
 
+@dataclass
+class Coverage:
+    level: str
+    items: int
+    ok: int
```
SurajBDeore, line 38: please align the column

lobster/report.py
```diff
@@ -193,7 +200,39 @@ def load_report(self, filename):
...
+                    self.mh.error(loc, "%s is not an array" % rkey)
+            elif rkey == "policy":
+                if not isinstance(data[rkey], dict):
+                    self.mh.error(loc, "policy is not an object")
+            elif rkey == "version":
+                if not isinstance(data[rkey], int):
+                    self.mh.error(loc, "version is not an integer")
+            else:
+                if not isinstance(data[rkey], str):
+                    self.mh.error(loc, "%s is not a string" % rkey)
```
SurajBDeore, line 238: I would suggest instead of checking in if elif elase block we can pass dict which contain key and data type and in single if block we can check data type is correct or not if not then we will though an error like below 
{"schema : str,
"version" : int,}
if not isntance(data[key], type)
    self.mh.error((loc, "%s is not a type" % type)

lobster/report.py
kedarnn: Done

lobster/report.py
kedarnn: Done

lobster/report.py
kedarnn: Made the change

mugdhadhole1 (dismissed):

lobster/report.py
```diff
@@ -193,7 +208,55 @@ def load_report(self, filename):
...
+
+    def validate_indicated_schema(self, data, loc):
+        """
+        Function validates the schema and version.
+        Parameters
+        ----------
+        data - contents of lobster json file.
+        loc  - location from where the error was raised.
+
+        Returns - None
```
mugdhadhole1:
```suggestion
        Returns - Nothing
```

lobster/report.py
```diff
@@ -193,7 +208,55 @@ def load_report(self, filename):
...
+    def validate_basic_structure_of_lobster_file(self, data, loc):
+        """
+        Function validates the basic structure of lobster file. All the first level
+        keys of the lobster json file are validated here.
+        Parameters
+        ----------
+        data - contents of lobster json file.
+        loc  - location from where the error was raised.
+
+        Returns - None
```
mugdhadhole1:
```suggestion
        Returns - Nothing
```

PR #125, opened 2024-11-15
system-test for `lobster-trlc`

SurajBDeore:
Add one requirements-based test for `lobster-trlc`.
Added small Python script to execute system tests for `trlc-lobster`. This script compares the actual output against the expected output.


tests-system/lobster-python/basic.output
```diff
@@ -10,7 +10,7 @@ Written output for 3 items to basic.lobster
         "line": 5,
         "column": null
       },
-      "name": "basic.trlc_reference",
```
phiwuu, line 13: don't merge this change

phiwuu (changes requested):

tests-system/lobster-trlc/rbt-output-file/default-scenario/expected-output/trlc.lobster
```diff
@@ -0,0 +1,25 @@
+{
+  "data": [
+    {
+      "tag": "req test_default.goodname",
+      "location": {
+        "kind": "file",
+        "file": "./default_file.trlc",
```
phiwuu, line 7: There is an issue here: on Windows the paths use `\\` instead of `/`. So this test will only succeed on Linux.

phiwuu (dismissed):

phiwuu (approved):

PR #129, opened 2024-11-22
Feature/lobster version flag for tools

kedarnn:


lobster/version.py
```diff
@@ -25,3 +27,42 @@
...
+    """
+    This decorator function is used on function wherever we are parsing
+    the command line arguments which then adds a version argument to the function.
+    If a version flag is passed to the command line arguments then the LOBSTER version
+    is printed.
+    Parameters
+    ----------
+    obj - obj can be an ArgumentParser object or a Function object.
+
+    Returns - None
```
mugdhadhole1, line 42:
```suggestion
    Returns - Nothing
```

mugdhadhole1 (changes requested):
`-v` and `--version`
command runs but the tool version is not printed as output

mugdhadhole1 (changes requested):
While I check the version for Lobster-cpp tool, it does show the version, but it asks for mandatory arguments as input to the tool

kedarnn:
Have added support for all the missing tools.

phiwuu (commented):
Please update the `CHANGELOG.md`

mugdhadhole1 (approved):

PR #131, opened 2024-11-28
Print all "duplicate definition" errors instead of only the first

phiwuu:
Function `lobster_read` prints all "duplicate definition" errors, and calls `mh.error` (with `fatal=True`) only for the last duplicate definition.

This helps the users to get an overview of all the duplicate definitions in their input files.
Previously the function raised an error already on the first occurrence, and the user had to fix that one, and call the L.O.B.S.T.E.R tools over and over again, until all errors were fixed.

**Note:**
`lobster-codebeamer` is affected when the parameter `--import-tagged` is used, as the tool then reads a lobster file to figure out which items to download from codebeamer.
Apart from that, only `lobster-report` is affected by the change.

`lobster-ci-report` is not affected, because it uses a different function to read `*.lobster` files.


tests-system/lobster-core/ci_report/rbt-ci-input-duplicate-definition/multiple-duplicates/expected-output/stdout.txt
```diff
@@ -0,0 +1,4 @@
+DuplicateDefinitionTest.py:2: lobster error: duplicate definition of hello DuplicateDefinitionTest, previously defined at DuplicateDefinitionTest.py:1
```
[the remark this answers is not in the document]
phiwuu: I have removed the test for `lobster-ci-report`. It was out of place in this pull request anyway.

lobster/io.py
```diff
@@ -124,10 +125,18 @@ def lobster_read(mh, filename, level, items, source_info=None):
...
-            items[item.tag.key()] = item
+                # 'duplicate definition' errors are fatal, but the user wants to see all
+                # of them. So store the affected items in a list first, and create
+                # errors later.
+                duplicate_items.append(item)
+            else:
+                items[item.tag.key()] = item
+
+    if duplicate_items:
+        for i, item in enumerate(duplicate_items, start=1):
```
kedarnn, line 136: Suggestion - Instead of using 'i' can a meaningful name be added here?

phiwuu: Done.

SurajBDeore (approved):

PR #134, opened 2024-12-02
Feature/html report UI upgrade

kedarnn:
This feature adds the following updates to the HTML report.

1. Filter items by status (Ok, Missing, Partial, Warning, Justified)
2. Hide/Unhide Issues.
3. Search in issues.


phiwuu:
The integration test is failing because the file `html_report.css` is not included in the wheel file. The integration tests builds each wheel file, installs them locally, and runs the lobster tools from there. Now running `lobster-html-report` fails because the `html_report.css` has not been packaged into the wheel file.

PR #148, opened 2024-12-10
selenium system test into CI for html_report

SurajBDeore:


kedarnn (commented):
Adding the review comments that we discussed for further reference.
1. Makefile - Make the execution on a folder instead of a single file.
2. Create a class for grouping the tests.

kedarnn (dismissed):

kedarnn (approved):

PR #149, opened 2024-12-10
Feature/use run tool tests py for all system tests

DiFerMa:
- CMD errout.txt is compared against expected-output
- Coverage command is run inside the tool
- lobster-python/system-tests was added to run like lobster-trlc


phiwuu:
Side remark: this is not a feature, but something internal. So the branch name should not be named "Feature/..." but "internal/..."

Makefile
```diff
@@ -58,7 +58,8 @@ integration-tests: packages
 
 system-tests:
 	mkdir -p docs
-	make -B -C tests-system/lobster-trlc
+	make -B -C tests-system TOOL=lobster-trlc
+#	make -B -C tests-system TOOL=lobster-python
 	make -B -C tests-system/lobster-json
 	make -B -C tests-system/lobster-python
```
phiwuu, line 64: Please apply the same approach to `lobster-json` and `lobster-python`, too.

tests-system/lobster-python/rbt-output-file/default-scenario/expected-output/basic.lobster
```diff
@@ -0,0 +1,65 @@
+Written output for 3 items to basic.lobster
+==========
```
phiwuu, line 2: These two lines must be removed when using `run_tool_tests.py`. Instead, the first line must be moved into `std.out.txt`.

tests-system/lobster-python/rbt-output-file/default-scenario/expected-output/stdout.txt
```diff
@@ -0,0 +1,64 @@
+{
```
phiwuu, line 1: Please use `lobster-python --out=basic.lobster` such that it creates a dedicated output file.
The system tests shall consist of (at least) two files for the expected output:
- `stdout.txt` for the expected command line output
- `basic.lobster` for the expected lobster file

tests-system/lobster-python/rbt-output-file/default-scenario/input/args.txt
```diff
@@ -0,0 +1,4 @@
+.
```
phiwuu, line 1: What is the requirement that the test shall verify?
I guess you want to add a test to show the refactoring is working, but the test itself is of minor importance in the context of this refactoring.
Please rename the folder to `rbt-dummy-requirement` in that case, instead of `rbt-output-file`.

tests-system/run_tool_tests.py
```diff
@@ -83,9 +84,9 @@ def _get_args(self) -> List[str]:
         test) from the corresponding test setup file"""
         file = join(self._input_folder, self._ARGS_FILE_NAME)
         with open(file, "r", encoding="UTF-8") as file:
-            return file.readlines()
+            return [argument.strip() for argument in file.readlines()]
 
-    def get_expected_cmd_output(self) -> str:
+    def get_expected_stdout_output(self) -> str:
```
phiwuu, line 89: The name could be simplified to `get_expected_stdout`

tests-system/run_tool_tests.py
```diff
@@ -95,6 +96,16 @@ def get_expected_cmd_output(self) -> str:
         with open(cmd_file, "r", encoding="UTF-8") as file:
             return file.read()
 
+    def get_expected_stderr_output(self) -> str:
+        """Reads the expected command line output (for stderr) from the corresponding
+        test setup file"""
+        cmd_file = join(
```
phiwuu, line 102: `cmd_file` could be named `stderr_file` here

phiwuu (changes requested):

phiwuu (approved):

Issue #87, opened 2024-10-08
`lobster-codebeamer` shall support query strings


phiwuu:
Instead of providing a codebeamer query ID to `lobster-codebeamer`, I want to specify the query string (in cbQL) directly. The advantage is that, I have code ownership over the query string in the git repository of the lobster CI job. This is not possible with the content of a query ID, as there is no concept like "code ownership" or "review" in codebeamer admin pages.

PR #158, opened 2024-12-16
Missing changes for the lobster-codebeamer supports query string argument

mugdhadhole1:
Implemented the missing changes the lobster-codebeamer supports query string argument

GitHub issue: #87

### DIFF
--- a/lobster/tools/codebeamer/codebeamer.py
+++ b/lobster/tools/codebeamer/codebeamer.py
@@ -170,8 +170,10 @@ def get_query(mh, cb_config, query):
                         page_id,
                         cb_config["page_size"]))
         elif isinstance(query, str):
-            url = ("%s/items/query?queryString=%s" %
+            url = ("%s/items/query?page=%u&pageSize=%u&queryString=%s" %
                     (cb_config["base"],
+                        page_id,
+                        cb_config["page_size"],
                         query))
         data = query_cb_single(cb_config, url)
         assert len(data) == 4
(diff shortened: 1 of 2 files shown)


lobster/tools/codebeamer/codebeamer.py
```diff
@@ -535,18 +537,14 @@ def main():
...
-                query = int(options.import_query)
-                if query < 1:
-                    ap.error("query_string must be a positive integer")
-            elif isinstance(options.import_query, str):
-                if options.import_query.startswith("-"):
-                    ap.error("query_string must be a positive integer"
-                                     " or valid string")
+            if isinstance(options.import_query, str):
+                if (options.import_query.startswith("-") and
+                    options.import_query[1].isdigit()):
```
kedarnn, line 542:
```suggestion
                    options.import_query[1:].isdigit()):
```
Need to check the whole string after `-`

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -535,18 +537,14 @@ def main():
...
+                    ap.error("import-query must be a positive integer")
+                elif options.import_query.startswith("-"):
+                    ap.error("import-query must be a valid cbQL query")
                 elif options.import_query.isdigit():
                     query = int(options.import_query)
-                    if query < 1:
-                        ap.error("query_string must be a positive integer")
                 else:
```
SurajBDeore, line 548: I think this statement is not needed by default it is string. I would suggest please try to test without this statement

SurajBDeore (approved):

kedarnn (approved):

PR #164, opened 2024-12-23
Added feature: query string support through config files

mugdhadhole1:
GitHub Issue: #87
Query string (CBQL query) can be now passed as a parameter in config file to the lobster codebeamer tool

### DIFF
--- a/lobster/tools/codebeamer/codebeamer.py
+++ b/lobster/tools/codebeamer/codebeamer.py
@@ -534,16 +534,18 @@ def main():
 
     elif cb_config.get("import_query"):
         try:
-            if isinstance(cb_config["import_query"], int):
-                cb_config["import_query"] = int(cb_config["import_query"])
-                if cb_config["import_query"] < 0:
-                    ap.error("import-query must be a positive integer")
+            if (isinstance(cb_config["import_query"], int) and
+                cb_config["import_query"] < 0):
+                ap.error("import-query must be a positive integer")
             elif isinstance(cb_config["import_query"], str):
-                if (cb_config["import_query"].startswith("-") and
-                    cb_config["import_query"][1:].isdigit()):
-                    ap.error("import-query must be a positive integer")
-                elif cb_config["import_query"].startswith("-"):
-                    ap.error("import-query must be a valid cbQL query")
+                if cb_config["import_query"].startswith("-"):
+                    if cb_config["import_query"][1:].isdigit():
+                        ap.error("import-query must be a positive integer")
+                    else:
+                        ap.error("import-query must be a valid cbQL query")
+                elif cb_config["import_query"][1:].isdigit():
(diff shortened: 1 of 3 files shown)


lobster/tools/codebeamer/codebeamer.py
```diff
@@ -522,17 +534,24 @@ def main():
...
+            if isinstance(cb_config["import_query"], int):
+                cb_config["import_query"] = int(cb_config["import_query"])
+                if cb_config["import_query"] < 0:
+                    ap.error("import-query must be a positive integer")
+            elif isinstance(cb_config["import_query"], str):
+                if (cb_config["import_query"].startswith("-") and
+                    cb_config["import_query"][1:].isdigit()):
+                    ap.error("import-query must be a positive integer")
+                elif cb_config["import_query"].startswith("-"):
+                    ap.error("import-query must be a valid cbQL query")
```
kedarnn, line 546:
```suggestion
            if ((isinstance(cb_config["import_query"], int) and int(cb_config["import_query"])<0)
                 or (isinstance(cb_config["import_query"], str) and cb_config["import_query"].startswith("-"))):
                 ap.error("import-query must be a valid codebeamer report id or a valid cbQL query ")
     
```

mugdhadhole1: This code won't work as it is
What if the user provides input like below
`import_query: "271679129"`

This input is then considered as a "query string"

kedarnn: okay got it.

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -522,17 +534,25 @@ def main():
...
-            sys.exit("lobster-codebeamer: Query ID must be a positive integer.")
+            if (isinstance(cb_config["import_query"], int) and
+                cb_config["import_query"] < 0):
+                ap.error("import-query must be a positive integer")
+            elif isinstance(cb_config["import_query"], str):
+                if (cb_config["import_query"].startswith("-") and
+                    cb_config["import_query"][1:].isdigit()):
+                    ap.error("import-query must be a positive integer")
+                elif cb_config["import_query"].startswith("-"):
+                    ap.error("import-query must be a valid cbQL query")
```
kedarnn, line 545:
```suggestion
                if cb_config["import_query"].startswith("-"):
                    if cb_config["import_query"][1:].isdigit()):
                       ap.error("import-query must be a positive integer")
                    else:
                       ap.error("import-query must be a valid cbQL query")
```

You can simplify this one.

mugdhadhole1: Here, this code also does not work if the user input is as below,
`import_query: "271679129"`

This input is then considered as a "query string"

we must convert the ID provided as string to int

SurajBDeore (approved):

PR #179, opened 2025-01-20
Fix assert

NicoletaCroitoru:
Fix assert

### DIFF
--- a/lobster/tools/core/online_report/online_report.py
+++ b/lobster/tools/core/online_report/online_report.py
@@ -218,29 +218,29 @@ def main():
 
     for item in report.items.values():
         if isinstance(item.location, File_Reference):
-            assert os.path.isdir(item.location.filename) or \
-                os.path.isfile(item.location.filename)
+            if os.path.isdir(item.location.filename) or \
+               os.path.isfile(item.location.filename):
 
-            rel_path_from_root = os.path.relpath(item.location.filename,
-                                                 repo_root)
-            # pylint: disable=possibly-used-before-assignment
-            actual_repo = gh_root
-            actual_sha  = options.commit
-            actual_path = rel_path_from_root
-            # pylint: disable=consider-using-dict-items
-            for prefix in gh_submodule_roots:
-                if path_starts_with_subpath(rel_path_from_root, prefix):
-                    actual_repo = gh_submodule_roots[prefix]
-                    actual_sha  = gh_submodule_sha[prefix]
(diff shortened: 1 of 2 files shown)


lobster/tools/core/online_report/online_report.py
```diff
@@ -218,29 +218,29 @@ def main():
 
     for item in report.items.values():
         if isinstance(item.location, File_Reference):
-            assert os.path.isdir(item.location.filename) or \
```
phiwuu, line 221: I don't understand this fix.
The idea is the following:
If `item.location` is of type `File_Reference`, then it must be a directory or a file (i.e. not a symlink).
And afterwards the rest of the code can be executed.

Which issue is being fixed by removing this assertion?

lobster/tools/cpp/cpp.py
```diff
@@ -161,15 +161,15 @@ def main():
             tag = Tracing_Tag("cpp", function_uid)
             loc = File_Reference(filename, line_nr)
 
-            assert tag.key() not in db
```
phiwuu, line 164: Here the idea of the assertion is to ensure that `db[tag.key()]` will not be overwritten. If it already exists in `db`, then that duplicate definition must be detected. We could raise a better user-friendly error message of course.

phiwuu (changes requested):

phiwuu:
Please always add an entry to `CHANGELOG.md` when changing something that affects the behavior of one of the tools.

lobster/tools/cpp/cpp.py
NicoletaCroitoru: In case there is a duplicate definition, lobster-cpp will crash, and the idea is just to go next without doing anything in this particular case by just using an "if" statement instead of "assert". 
Same explanation for lobster-online-repost, lets not simply crash the tool and do nothing.
Suggestion: Crash a tool if something goes wrong is not ok, any tool must recover from any kind of errors and never crash.

NicoletaCroitoru: image

NicoletaCroitoru: Use case: same file into two different folders.

phiwuu: I agree that a tool must never crash. But then please raise an exception instead of silently ignoring the error.

phiwuu (changes requested):

PR #180, opened 2025-01-20
Get full path of files found in rv.stdout

NicoletaCroitoru:
Get full path of files found in rv.stdout

### DIFF
--- a/lobster/tools/cpp/cpp.py
+++ b/lobster/tools/cpp/cpp.py
@@ -153,6 +153,7 @@ def main():
         match = re.match(RE_NOTAGS, line)
         if match:
             filename, line_nr, kind, function_name = match.groups()
+            filename = next((item for item in file_list if filename in item), None)
             filename = os.path.relpath(filename, prefix)
             line_nr = int(line_nr)
             function_uid = "%s:%s:%u" % (os.path.basename(filename),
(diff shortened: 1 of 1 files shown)


phiwuu (changes requested):
1. Please always add an entry to `CHANGELOG.md` when changing something that affects the behavior of one of the tools.
2. The integration test for `lobster-cpp` is failing.

lobster/tools/cpp/cpp.py
```diff
@@ -153,6 +153,7 @@ def main():
         match = re.match(RE_NOTAGS, line)
         if match:
             filename, line_nr, kind, function_name = match.groups()
+            filename = next((item for item in file_list if filename in item), None)
             filename = os.path.relpath(filename, prefix)
```
phiwuu, line 157: This line returns the relative path, but your pull request title indicates to use the full path. What do you intend to achieve?

NicoletaCroitoru: Use case: some lines from rv.stdout contains relative path instead of absolute path and because of this filename = os.path.relpath(filename, prefix) will do nothing and a fix to get the absolute path is needed.

phiwuu:
Implemented with https://github.com/bmw-software-engineering/lobster/pull/268, so I am closing this pull request.

PR #181, opened 2025-01-21
Allow to see all errors detected.

NicoletaCroitoru:
Allow to see all errors detected, so they can be fixed.

### DIFF
--- a/lobster/tools/cpp/cpp.py
+++ b/lobster/tools/cpp/cpp.py
@@ -142,7 +142,6 @@ def main():
             print(rv.stdout)
         print()
         print(rv.stderr)
-        return 1
 
     db = {}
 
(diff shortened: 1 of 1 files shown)


lobster/tools/cpp/cpp.py
```diff
@@ -142,7 +142,6 @@ def main():
             print(rv.stdout)
         print()
         print(rv.stderr)
-        return 1
```
phiwuu, line 145: If you remove the `return 1` here, then you have to add it somewhere else. Otherwise the tool may return with zero instead.

lobster/tools/cpp/cpp.py
```diff
@@ -219,7 +218,6 @@ def main():
 
         print("could not parse line")
         print(">", line)
-        return 1
```
phiwuu, line 222: If you remove the `return 1` here, then you have to add it somewhere else. Otherwise the tool may return with zero instead.

phiwuu (changes requested):

phiwuu:
Implemented in https://github.com/bmw-software-engineering/lobster/pull/267, so I will close this PR.

PR #200, opened 2025-02-11
List test case folders in LOBSTER report

phiwuu:
Use the folders of the test cases in LOBSTER reports, instead of the test suite folder.


phiwuu:
Will be replaced by the new approach which is about to be implemented with https://github.com/bmw-software-engineering/lobster/pull/215

PR #215, opened 2025-02-21
System Test Framework based on `unittest`

phiwuu:
Implement a new version of the system test framework.
It is based on Python's `unittest` framework.

Add requirements for the system test framework.
Add a `README.md` to explain these requirements and to give context.


tests-system/testrunner.py
```diff
@@ -0,0 +1,110 @@
...
+    - The test runner can be configured with a working directory that is used as the
+      current working directory when the tool is run.
+    - The test runner can run the tool under test and return the result as a
+      subprocess.CompletedProcess object.
+    """
+
+    def __init__(self, tool_name: str, working_dir: Path):
+        """Constructor
+
+        :param tool_name: Name of the LOBSTER tool. This must be qual to the name of the
```
kedarnn, line 28:
```suggestion
        :param tool_name: Name of the LOBSTER tool. This must be equal to the name of the
```

tests-system/systemtestcase.py
```diff
@@ -0,0 +1,24 @@
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from typing import List
+from unittest import TestCase
+
+
+class SystemTestCase(TestCase):
+    def __init__(self, methodName):
+        super().__init__(methodName)
+        self.maxDiff = None  # pylint: disable=invalid-name
```
kedarnn, line 10: What is the purpose of this variable?

tests-system/testrunner.py
phiwuu: Fixed

tests-system/systemtestcase.py
phiwuu: See https://docs.python.org/3/library/unittest.html#unittest.TestCase.maxDiff

Sebastian-Koerner (approved):

PR #224, opened 2025-03-03
Support empty codebeamer queries

phiwuu:
Modify `lobster-codebeamer` such that it creates an output file even if the CB query returned no items.

Added requirements.

### DIFF
--- a/lobster/tools/codebeamer/codebeamer.py
+++ b/lobster/tools/codebeamer/codebeamer.py
@@ -51,6 +51,10 @@ TOKEN = 'token'
 REFERENCES = 'references'
 
 
+class CodebeamerError(Exception):
+    pass
+
+
 class SupportedConfigKeys(Enum):
     """Helper class to define supported configuration keys."""
     IMPORT_TAGGED = "import_tagged"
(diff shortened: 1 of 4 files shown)


lobster/tools/codebeamer/codebeamer.py
```diff
@@ -186,22 +186,22 @@ def get_query(mh, cb_config, query):
...
-            sys.exit(1)
+            print(f"You can try to access '{url}' manually to check.")
 
         assert page_id == data["page"]
         if page_id == 1:
             total_items = data["total"]
         else:
             assert total_items == data["total"]
 
-        if query is not None and isinstance(query, int):
```
phiwuu, line 201: Note: I have removed the condition `query is not None and` because the function already contains `assert isinstance(query, (int, str))` at the beginning.

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -541,14 +541,16 @@ def main():
                 except ValueError:
                     pass
 
-    elif cb_config.get("import_query"):
+    elif cb_config.get("import_query") is not None:
```
phiwuu, line 544: Note: If there was an empty string given as value for `import_query`, like this:
```yaml
import_query: ""
```
then `cb_config.get("import_query")` evaluates to `False` and the `elif` block is not entered, and hence the checks are not being performed. This is wrong. The checks must be performed whenever there is any kind of value given, so we have to check for `not None` instead.

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -186,22 +190,25 @@ def get_query(mh, cb_config, query):
...
             print("This query doesn't generate items. Please check:")
             print(" * is the number actually correct?")
             print(" * do you have permissions to access it?")
-            print("You can try to access %s manually to check" % url)
-            sys.exit(1)
+            print(f"You can try to access '{url}' manually to check.")
+
+        if page_id != data["page"]:
+            raise CodebeamerError(f"Page mismatch in query result: expected page "
+                                  f"{page_id} from codebeamer, but got {data['page']}")
```
phiwuu, line 201: This used to be an assertion, ` assert page_id == data["page"]`
This is no good API design. So we are raising an exception now.

I had to refactor this part because the unit test `test_get_query_with_invalid_data` was testing that `sys.exit(1)` gets called. But in fact the test prepares a setup where codebeamer replies with inconsistent data: it returns zero items in the first page request, but the meta-data says it would return one item. In this case `CodebeamerError` must be raised.

tests-unit/lobster-codebeamer/test_codebeamer.py
```diff
@@ -20,6 +20,13 @@
 
 
 class QueryCodebeamerTest(unittest.TestCase):
+    def setUp(self):
+        self._mock_cb_config = {
+            "root" : "http://some.codebeamer.server",
+            "base": "protocol://base.api.url/api/v3",
+            "page_size": 10
+        }
```
phiwuu, line 28: This part was duplicated in all the unit tests, so I pulled into the `setUp` method.

PR #225, opened 2025-03-03
Use `SupportedConfigKeys` instead of strings

phiwuu:


lobster/tool.py
```diff
@@ -124,11 +124,12 @@ def process_commandline_options(
         config = self.load_yaml_config(options.config)
 
         if not options.out:
-            options.out = config.get("out")
-        options.inputs_from_file = config.get("inputs_from_file")
-        options.inputs = config.get("inputs", [])
-        options.traverse_bazel_dirs = config.get("traverse_bazel_dirs", False)
-        options.single = config.get("single", False)
+            options.out = config.get(SupportedConfigKeys.OUT.value)
+        options.inputs_from_file = config.get(SupportedConfigKeys.INPUT_FROM_FILE.value)
```
kedarnn, line 128: `INPUT_FROM_FILE` is just a `String` and it should be just `SupportedConfigKeys.INPUT_FROM_FILE` instead of `SupportedConfigKeys.INPUT_FROM_FILE.value`.

kedarnn (approved):

PR #231, opened 2025-03-10
Yaml config for LOBSTER Json tool

kedarnn:
The LOBSTER Json command line arguments are now converted to Yaml config file arguments. This PR also contains modification for some common functions in LOBSTER_Tool class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,8 @@
 
 ### 0.11.1-dev
 
-* `lobster-json` - now all the commandline arguments except `--out` are replaced by 
-  Yaml config file which means now only `--out` and `--config` are supported command 
-  line arguments.
+* `lobster-json` - All the command-line arguments are moved to Yaml based config 
+  file. `--config` and `--out` command-line arguments are still supported.
 
 ### 0.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@
 
 ### 0.11.1-dev
 
-
+* `lobster-json` - now all the commandline arguments except `--out` are replaced by 
+  Yaml config file which means now only `--out` and `--config` are supported command 
+  line arguments.
 
 ### 0.11.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,8 @@
 
 ### 0.11.1-dev
 
-* `lobster-json` - All the command-line arguments are moved to Yaml based config 
-  file. `--config` and `--out` command-line arguments are still supported.
+* `lobster-json` - All command-line arguments except  `--config` and `--out` are 
+  moved to Yaml based config file. `--config` and `--out` command-line arguments are still supported.
 
 ### 0.11.0
 
--- a/lobster/tool.py
+++ b/lobster/tool.py
@@ -25,10 +25,10 @@ import multiprocessing
 from abc import ABCMeta, abstractmethod
 from functools import partial
 from typing import List, Union, Tuple
-from enum import Enum
 import yaml
 from lobster.version import FULL_NAME, get_version
-from lobster.errors import Message_Handler
+from lobster.errors import (Message_Handler, UnsupportedParametersError,
+                            MandatoryParametersError)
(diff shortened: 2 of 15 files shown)


mugdhadhole1 (changes requested):
You should also make changes in integration test where lobster-json is tested
The CI job has failed for lobster-json integration test

CHANGELOG.md
```diff
@@ -5,7 +5,9 @@
 
 ### 0.11.1-dev
 
-
+* `lobster-json` - now all the commandline arguments except `--out` are replaced by 
+  Yaml config file which means now only `--out` and `--config` are supported command 
+  line arguments.
```
SurajBDeore, line 11: This entry is very confusing.

lobster/tool.py
```diff
@@ -73,63 +100,98 @@ def __init__(self, name, description, extensions, official):
...
-
-        self.g_common = self.ap.add_argument_group("common options")
-        self.g_tool   = self.ap.add_argument_group("tool specific options")
-
-        self.g_common.add_argument(
-            "--config",
-            default = None,
-            help=("Path to YAML file with arguments, "
-                          "supported references: '%s'" %
-                          ', '.join(SupportedConfigKeys.as_set())),
```
phiwuu, line 86: `--config` is be a parameter that **every** lobster tool will need. I guess you decided to remove it because the help message will be different for each tool, didn't you?
Do we want to provide documentation of `SupportedConfigKeys` here at all? Imagine the number of config parameter increases. Maybe it is more helpful to document them in a `README.md` file instead. Then `--config` could remain in class `LOBSTER_Tool`.

kedarnn: Yes the help message would be different for every tool but yes you are correct, the `--config` parameter can be moved to `LOBSTER_Tool` class.
I feel that it is very helpful for the user to check the configuration parameters using help flag instead of going to the `README.md` and then check it. If we take example of `git` or `pip` then they have huge list of parameters for some of their commands. 
Hence, in my view we should keep all the config parameters in help command to make it handy for the user.

tests-system/lobster-json/lobsterjsontestrunner.py
phiwuu, line 1: The changes in this file look good.

lobster/tool.py
```diff
@@ -25,31 +25,57 @@
...
 
 BUG_URL = "https://github.com/bmw-software-engineering/lobster/issues"
 
 
-class SupportedConfigKeys(Enum):
+class SupportedCommonConfigKeys:
     """Helper class to define supported configuration keys."""
     INPUT_FROM_FILE     = "inputs_from_file"
     TRAVERSE_BAZEL_DIRS = "traverse_bazel_dirs"
     SINGLE              = "single"
```
phiwuu, line 43: For now, only `lobster-json` supports multithreading. In the future all lobster tools shall support that. But for now, please move "single" out of the `LOBSTER_Tool` class and into `lobster-json`.

lobster/tool.py
```diff
@@ -73,63 +100,98 @@ def __init__(self, name, description, extensions, official):
...
+        mandatory_parameters = self.get_mandatory_parameters()
+        config_keys = data.keys()
+
+        try:
+            if not all(param in config_keys for param in mandatory_parameters):
+                raise MandatoryParametersError(
+                    mandatory_parameters
+                )
+        except MandatoryParametersError as e:
+            print(e)
```
phiwuu, line 175: please use `self.mh.error` to collect and print errors.

CHANGELOG.md
```diff
@@ -5,7 +5,8 @@
 
 ### 0.11.1-dev
 
-
+* `lobster-json` - All the command-line arguments are moved to Yaml based config 
+  file. `--config` and `--out` command-line arguments are still supported.
```
phiwuu, line 9:
```suggestion
* `lobster-json` - All command-line arguments except `--config` and `--out`
  are moved to Yaml based config file.
```

phiwuu (changes requested):

phiwuu (approved):

PR #237, opened 2025-03-17
Migration - lobster-report system tests to new STF

mugdhadhole1:
Migrated following lobster-report system tests to STF 

- Missing
- Justified
- Ok - made changes in OK status system test


tests-system/lobster-html-report/rbt-input-file-existance/is-file/input/is_actually_html.lobster
kedarnn, line 1: Remove this file

tests-system/lobster-online-report/rbt-valid-flow/valid-scenario/expected-output/expected-output.lobster
kedarnn, line 1: Remove this file

tests-system/lobster-report/run_report_tests.py
```diff
@@ -0,0 +1,24 @@
...
+        self._test_runner.cmd_args.out = out_file
+        self._test_runner.cmd_args.lobster_config = config_file
+        self._test_runner.declare_output_file(self._data_directory / out_file)
+
+        completed_process = self._test_runner.run_tool_test()
+        asserter = Asserter(self, completed_process, self._test_runner)
+        asserter.assertNoStdErrText()
+        asserter.assertNoStdOutText()
+        asserter.assertExitCode(0)
+        asserter.assertOutputFiles()
```
kedarnn, line 24: This setup mandates that all the system tests from lobster-report shall have common assertions and compulsory need of input files. Just think that now we have very limited tests but in future there will be many tests having specific assertions then this will not work. Hence, keeping the assertions and inputs tests specific would be more feasible.

mugdhadhole1: I get linting errors which state that the code is repeated i.e, duplicate lines of code in multiple files when I run the command
`make lint-system-tests`
To avoid this, I have combined the similar lines of code.

Could you suggest any way to avoid these duplication of code linting error?

kedarnn (approved):

PR #243, opened 2025-03-18
Fix publishment of docs

phiwuu:
On the one hand it is necessary to run the system and unit tests in order to create the branch coverage report. The system tests require that graphviz is NOT installed, otherwise they fail.

On the other hand, the LOBSTER reports shall show the visualization of the tracing policy, so graphviz must be available for their generation.

The `docs.yml` workflow creates all reports and publishes them to GitHub pages. Hence the order of steps matters:
1. generate the branch coverage reports
2. install graphviz
3. generate the LOBSTER traceability reports

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -128,10 +128,12 @@ test-all: integration-tests system-tests unit-tests
 	util/check_local_modifications.sh
 
 docs:
-	rm -rf docs
 	mkdir -p docs
 	@-make tracing
 	@-make tracing-stf
+
+clean-docs:
+	rm -rf docs
 
 tracing:
 	@mkdir -p docs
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -128,11 +128,13 @@ test-all: integration-tests system-tests unit-tests
 	util/check_local_modifications.sh
 
 docs:
-	rm -rf docs
```
phiwuu, line 131: Remember: the order of steps in `docs.yml` is as follows:
1. make coverage
2. install graphviz
3. make docs

So the last step `make docs` must not delete the `docs` folder, because `docs.yml` has already moved the folder of the coverage report to `docs`.

That's why I had to remove this line from the `Makefile`.

Sebastian-Koerner (approved):

PR #257, opened 2025-03-27
Precise summary message of `lobster-online-report`

phiwuu:
Write the summary message of `lobster-online-report` more precisely
so that it becomes clear whether the input file has been modified,
or whether a new output file has been created.

### DIFF
--- a/lobster/tools/core/online_report/online_report.py
+++ b/lobster/tools/core/online_report/online_report.py
@@ -233,10 +233,16 @@ def main():
                 commit=commit)
             item.location = loc
 
-    report.write_report(options.out if options.out else options.lobster_report)
-    print("LOBSTER report %s changed to use online references" %
-          options.out if options.out else options.lobster_report)
+    out_file = options.out if options.out else options.lobster_report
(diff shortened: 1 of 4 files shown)


lobster/tools/core/online_report/online_report.py
```diff
@@ -233,11 +233,17 @@ def main():
...
+    out_file = options.out if options.out else options.lobster_report
+    report.write_report(out_file)
+    print(get_summary(options.lobster_report, out_file))
     return 0
 
 
+def get_summary(in_file: str, out_file: str):
+    if in_file == out_file:
+        return f"LOBSTER report {in_file} modified to use online references."
+    return f"LOBSTER report {out_file} created, using online references."
```
kedarnn, line 245: `out_file` is already storing the file name based on if ... else condition. I think then this would not be needed.

phiwuu: Please note the difference in the wording of the message.

kedarnn (approved):

PR #258, opened 2025-03-27
Move authentication provision into separate function

phiwuu:
Extract the creation of the authentication data from `query_cb_single`
and move it into a separate function.

Clean up import statements.

Move class `BearerAuth` into separate file.

This change has no impact on features of `lobster-codebeamer`. It is only a refactoring.


Sebastian-Koerner (approved):

phiwuu:
Please ignore the branch name of this pull request. That was a mistake. This change has nothing to do with catching encoding errors.

PR #260, opened 2025-03-28
Added fix for reports generation

kedarnn:
lobster-trlc tool was modified to use Yaml configuration but the `tracing` target which is used to generate the `lobster-html-reports` for all the LOBSTER tools was not updated. Due to this issue the reports were not generated.
Also the tracing policy was modified and `Software Requirements` trace to `System Requirements` was removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -159,17 +159,25 @@ report.lobster-%: lobster/tools/lobster.conf \
 		--out=report.lobster
 	lobster-online-report report.lobster
 
-system_requirements.lobster-%: lobster/tools/requirements.rsl
-	$(eval TOOL_PATH := $(subst -,/,$*))   
-	lobster-trlc lobster/tools/$(TOOL_PATH) lobster/tools/requirements.rsl \
-	--config-file=lobster/tools/lobster-trlc-system.conf \
-	--out system_requirements.lobster
+system_requirements.lobster-%: TRLC_CONFIG = lobster/tools/lobster-trlc-system.conf
 
-software_requirements.lobster-%: lobster/tools/requirements.rsl
-	$(eval TOOL_PATH := $(subst -,/,$*))   
-	lobster-trlc lobster/tools/$(TOOL_PATH) lobster/tools/requirements.rsl \
-	--config-file=lobster/tools/lobster-trlc-software.conf \
-	--out software_requirements.lobster
+system_requirements.lobster-%:
+	$(eval TOOL_PATH := $(subst -,/,$*))
+	@echo "inputs: ['lobster/tools/requirements.rsl', 'lobster/tools/$(TOOL_PATH)']" > lobster/tools/config.yaml
+	@echo "trlc_config_file: $(TRLC_CONFIG)" >> lobster/tools/config.yaml
+	lobster-trlc --config=lobster/tools/config.yaml \
+	--out=system_requirements.lobster
+	rm lobster/tools/config.yaml
+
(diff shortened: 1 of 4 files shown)


lobster/tools/lobster.conf
```diff
@@ -4,6 +4,7 @@ requirements "System Requirements" {
 
 requirements "Software Requirements" {
   source: "software_requirements.lobster";
+  trace to: "System Requirements";
```
phiwuu, line 7: The trace to System Requirements have been removed on purpose, see https://github.com/bmw-software-engineering/lobster/commit/bb92390214c44aa1f179ab9e03bfac34a0cfaee6#diff-2095f3104c85e577e1b43ae04fe695a108e6ced4d81a4705f1aabac17db55500

It is a workaround to see the progress in qualifiying LOBSTER. There only the coverage from system requirements down to system tests is important, not the breakdown to software requirements. Currently LOBSTER can not give separate numbers for these two metrics.

phiwuu (changes requested):

Makefile
```diff
@@ -161,17 +161,25 @@ report.lobster-%: lobster/tools/lobster.conf \
...
-	$(eval TOOL_PATH := $(subst -,/,$*))   
-	lobster-trlc lobster/tools/$(TOOL_PATH) lobster/tools/requirements.rsl \
-	--config-file=lobster/tools/lobster-trlc-software.conf \
-	--out software_requirements.lobster
+system_requirements.lobster-%: TRLC_CONFIG = lobster/tools/lobster-trlc-system.conf
+
+system_requirements.lobster-%:
+	$(eval TOOL_PATH := $(subst -,/,$*))
+	@echo "inputs: ['lobster/tools/requirements.rsl', 'lobster/tools/$(TOOL_PATH)']" > lobster/tools/config.yaml
+	@echo "trlc_config_file: $(TRLC_CONFIG)" >> lobster/tools/config.yaml
```
phiwuu, line 169: Simply create a new file called `config.yaml`, store the values there, and add the file to the repository. Otherwise you should also delete the file after the make target finished.

Makefile
```diff
@@ -161,17 +161,25 @@ report.lobster-%: lobster/tools/lobster.conf \
...
+	lobster-trlc --config=lobster/tools/config.yaml \
+	--out=system_requirements.lobster
+	rm lobster/tools/config.yaml
+
+software_requirements.lobster-%: TRLC_CONFIG = lobster/tools/lobster-trlc-software.conf
+
+software_requirements.lobster-%:
+	$(eval TOOL_PATH := $(subst -,/,$*))
+	@echo "inputs: ['lobster/tools/requirements.rsl', 'lobster/tools/$(TOOL_PATH)']" > lobster/tools/config.yaml
+	@echo "trlc_config_file: $(TRLC_CONFIG)" >> lobster/tools/config.yaml
```
phiwuu, line 179: same as above

Makefile
kedarnn: I am deleting the file. Please see line no 172.

phiwuu (changes requested):
Please also fix `tracing-stf`

Makefile
kedarnn: As discussed, we will keep dynamic file generation for yaml config instead of having config files already in the repo. With this approach we don't need to create tool specific yaml config files which would have been the case otherwise.

Makefile
kedarnn: Added comment to the previous one.

mugdhadhole1 (approved):

phiwuu (approved):

PR #261, opened 2025-03-31
Feature: lobster-cpptest has test-name in tag

mugdhadhole1:
Earlier lobster-cpptest showed the fixture-name in tag in the HTML report.
This feature implementation will now show test-name instead of fixture name in the tag.


kedarnn:
I think this should be mentioned in the README as well.

kedarnn (approved):

PR #292, opened 2025-05-05
Added meta tag in html header

kedarnn:


lobster/tools/core/html_report/html_report_content.trlc
```diff
@@ -0,0 +1,10 @@
+package html_req
+import req
+
+req.System_Requirement Meta_Tag_In_Html_Report {
+    description = '''
+      IF the the generated HTML report is a valid HTML file
```
SurajBDeore, line 6: Typo: the the

SurajBDeore: suggestion: IF an HTML report is generated,
THEN the report shall include a <meta> tag
HAVING both the http-equiv and content attributes

if html report is generated then it means its valid file

tests-system/lobster-html-report/test_html_report.py
```diff
@@ -141,3 +141,12 @@ def test_git_hash_timestamp(self):
                         expected_time = datetime.fromtimestamp(
                             int(result.stdout.strip()), tz=timezone.utc)
                         self.assertIn(str(expected_time), text)
+
+    def test_meta_tag(self):
+        """Test to check the presence of meta tag in the HTML report."""
+        # lobster-trace: html_req.Meta_Tag_In_Html_Report
+        self._test_runner.run_tool_test()
+        self.driver.get(self.input_file)
+        element = self.driver.find_element(By.TAG_NAME, "meta")
```
SurajBDeore, line 150: element : meta_element

SurajBDeore (approved):

PR #298, opened 2025-05-15
Fix URL generation in `Codebeamer_Reference` class

phiwuu:
The `to_html` method appended an additional `/cb` to the root url,
leading to an incorrect url.
It also appends `?version=<version>` if a version is given.

Add type hints to the class constructor method.

Add unit tests.

The bug has been introduced with https://github.com/bmw-software-engineering/lobster/pull/201

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,10 @@
 
 ### 0.12.2-dev
 
-
+* `lobster-html-report`:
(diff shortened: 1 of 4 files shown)


kedarnn (approved):
Tested. The changes are working as expected.

phiwuu:
@kedarnn Thank you!

PR #306, opened 2025-05-20
SVG optimization code in HTML report

kedarnn:
The SVG are now initialized once and then reused wherever needed instead of declaring and then using it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@
 
 ### 0.12.3-dev
 
+* `lobster-html-report`
+  - The SVG's in the HTML report were rendered on every use. Now the SVGs
+    are rendered once and reused for at all the remaining occurences. 
+    This will reduce the overall HTML file size.
+
 * `Separate Coverage Reports`
   - Separate coverage reports for unit tests (`.coverage.unit`) and system tests (`.coverage.system`).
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,11 @@
 
 ### 0.12.3-dev
 
+* `lobster-html-report`
+  - The SVG's in the HTML report were rendered on every use. Now the SVGs
+    are rendered once and reused for at all the remaining occurences. 
+    This will reduce the overall HTML file size.
```
SurajBDeore, line 11: suggestion:
instead of this This will reduce the overall HTML file size - which improves loading performance and reduces redundancy.

lobster/tools/core/html_report/html_report_content.trlc
```diff
@@ -8,3 +8,11 @@ req.System_Requirement Meta_Tag_In_Html_Report {
       HAVING attributes "http-equiv" and "content".
     '''
 }
+
+req.System_Requirement HTML_Report_with_20k_records {
+    description = '''
+      IF a HTML report conatining approximate twenty thousand item is generated,
```
SurajBDeore, line 14: I am confused in below req.
- If the HTML report contains twenty thousand items, it means the tool has already executed successfully. At that point, we are checking it returns a zero exit code.

CHANGELOG.md
kedarnn: This is on purpose because this doesn't increase the performance except reducing the file size.

SurajBDeore (approved):

PR #310, opened 2025-05-21
Bug Fix: Reference Field Support in LOBSTER Codebeamer

SurajBDeore:
Fixed handling of references in LOBSTER Codebeamer to correctly process a list of reference names.

### DIFF
--- a/tests-system/lobster-codebeamer/test_valid_flow.py
+++ b/tests-system/lobster-codebeamer/test_valid_flow.py
@@ -2,7 +2,7 @@ import json
 from flask import Response
 from .lobster_codebeamer_system_test_case_base import (
     LobsterCodebeamerSystemTestCaseBase)
-from ..asserter import Asserter
+from .lobster_codebeamer_asserter import LobsterCodebeamerAsserter
 from .mock_server_setup import start_mock_server, get_mock_app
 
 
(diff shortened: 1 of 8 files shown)


tests-system/lobster-codebeamer/test_valid_flow.py
```diff
@@ -61,3 +61,49 @@ def test_valid_query_id(self):
...
+            ]
+        }
+
+        self.codebeamer_flask.responses = [
+            Response(json.dumps(response_data), status=200),
+        ]
+
+        completed_process = self._test_runner.run_tool_test()
+        asserter = Asserter(self, completed_process, self._test_runner)
+        self.assertIn("requirements to codebeamer.lobster", completed_process.stdout)
```
phiwuu, line 107: Is this a helpful assertion? It only makes sure that the message is printed, but the number could be completely wrong. How many items did it print? I recommend to write a `CodebeamerAsserter` class which can offer a generic implementation.

Take inspiration from: https://github.com/bmw-software-engineering/lobster/blob/2aab0d846a4a908df74ecda035cce76135bee89f/tests-system/lobster-json/lobsterjsonasserter.py#L10-L14

phiwuu (changes requested):

phiwuu:
Please provide a commit message.

kedarnn (changes requested):
Please add a CHANGELOG for this fix.

kedarnn (approved):

PR #311, opened 2025-05-22
Make mock item data unique

phiwuu:
At least one test of `lobster-codebeamer` shall verify that all properties of a codebeamer item are written in the correct
JSON keys of the corresponding LOBSTER item in the output file.

For example, codebeamer item with ID 123 and version 45 shall be written exactly like this, and an erroneously permutation of these two values shall be detected by the test.
This test rule applies to all properties of the codebeamer item.

Hence it is important that the mock data uses unique values for all properties among the set of all generated mock items. This way any permutation of any randomly picked values can easily be detected
by the system test.

This change implements the above requirement to have unique mock values.


tests-system/lobster-codebeamer/lobster_codebeamer_system_test_case_base.py
```diff
@@ -28,9 +28,12 @@ def set_config_file_data(self, retry_codes=None, num_retries=None):
         if num_retries is not None:
             cfg.num_request_retry = num_retries
 
-    @staticmethod
-    def create_mock_response_items(page: int, page_size: int, total: int):
+    def create_mock_response_items(self, page: int, page_size: int, total: int):
         """Create a mock response like codebeamer API paginated items."""
+        multiplicator = 100
```
kedarnn, line 33:
```suggestion
        MULTIPLICATOR = 100
```

kedarnn (dismissed):

kedarnn (approved):

PR #312, opened 2025-05-26
Use 0% coverage in case of zero items

phiwuu:
Set the coverage value to 0% if there are zero items in the tracing policy level.

Add unit test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@
 
 ### 0.12.3-dev
 
+* `lobster-report`
+  - If there are zero items in one level of the tracing policy, then this level now
+    shows a coverage of 0%.
+    Previously its coverage was 100%.
+    Note that the coverage ratio cannot be computed in a mathematical way if there
+    are zero items, because the formula requires to divide by the total number of items.
+    If this denominator is zero, obviously the division cannot be performed.
+    So neither 100% nor 0% makes sense, but choosing 0% is the safe decision
+    when the tool is used in a safety-critical context.
+    Having zero input items is probably not what the user intended to do,
+    and shall be notified about this empty input.
+    The user has a greater chance to detect this empty input if the report shows 0% coverage
+    compared to 100%, which indicates that everything is okay.
+
 * `lobster-cpp`
   * The file basename is used to construct the function UID.
     A counter is then appended to the basename to handle situations where files in
--- a/lobster/report.py
+++ b/lobster/report.py
@@ -102,12 +102,10 @@ class Report:
 
     def compute_coverage_for_items(self):
         for level_obj in self.coverage.values():
-            if level_obj.ok == level_obj.items:
-                level_obj.coverage = 100.0
+            if level_obj.items == 0:
+                level_obj.coverage = 0.0
             else:
-                level_obj.coverage = (
-                    float(level_obj.ok * 100) / float(level_obj.items)
-                )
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,20 @@
...
+* `lobster-report`
+  - If there are zero items in one level of the tracing policy, then this level now
+    shows a coverage of 0%.
+    Previously its coverage was 100%.
+    Note that the coverage ratio cannot be computed in a mathematical way if there
+    are zero items, because the formula requires to divide by the total number of items.
+    If this denominator is zero, obviously the division cannot be performed.
+    So neither 100% nor 0% makes sense, but chosing 0% is the safe decision
+    when the tool is used in a safety-critical context.
+    Having zero input items is probably not what the user intended to to,
```
mugdhadhole1, line 17:
```suggestion
    Having zero input items is probably not what the user intended to do,
```

tests-unit/lobster-report/test_report.py
```diff
@@ -1,4 +1,44 @@
-# please consider writing here your tests and set your tracing tags. 
-# This file will be used in the tracing report
-# if any other file or files are used, make the needed
-# adjsutments to tracing report target.
\ No newline at end of file
+import sys
```
SurajBDeore, line 1: The sys module is not used anywhere. Please remove it.

lobster/report.py
```diff
@@ -103,7 +103,7 @@ def resolve_references_for_items(self):
     def compute_coverage_for_items(self):
         for level_obj in self.coverage.values():
             if level_obj.ok == level_obj.items:
-                level_obj.coverage = 100.0
+                level_obj.coverage = 100.0 if level_obj.items else 0.0
```
SurajBDeore, line 106: I guess we don't need this condition anymore :   if level_obj.ok == level_obj.items:
Because if items != 0 and ok == items, the else condition will still calculate the coverage correctly
(ok * 100.0) / items becomes (items * 100.0) / items, which equals 100.0

CHANGELOG.md
```diff
@@ -5,6 +5,20 @@
...
+  - If there are zero items in one level of the tracing policy, then this level now
+    shows a coverage of 0%.
+    Previously its coverage was 100%.
+    Note that the coverage ratio cannot be computed in a mathematical way if there
+    are zero items, because the formula requires to divide by the total number of items.
+    If this denominator is zero, obviously the division cannot be performed.
+    So neither 100% nor 0% makes sense, but chosing 0% is the safe decision
+    when the tool is used in a safety-critical context.
+    Having zero input items is probably not what the user intended to do,
+    and shall be notified about this wrong input.
```
SurajBDeore, line 18: instead of wrong we can use invalid input

tests-unit/lobster-report/test_report.py
phiwuu: Good catch! It turns out that we don't have a lint target for unit tests!

lobster/report.py
phiwuu: Thanks, I have simplified the function like this:
```python
def compute_coverage_for_items(self):
    for level_obj in self.coverage.values():
        if level_obj.items == 0:
            level_obj.coverage = 0.0
        else:
            level_obj.coverage = float(level_obj.ok * 100) / float(level_obj.items)
```

SurajBDeore (approved):

PR #326, opened 2025-06-11
Use `realpath` when converting from file reference to github reference

phiwuu:


phiwuu:
> Changelog is missing.

Done.

kedarnn (approved):

PR #334, opened 2025-06-26
restrict the requirements regex

TannazVhdBMWExt:
with new regex multiple comment blocks consisting requirement tags would extracted from the main comment block

### DIFF
--- a/lobster/tools/cpptest/parser/test_case.py
+++ b/lobster/tools/cpptest/parser/test_case.py
@@ -168,17 +168,20 @@ class TestCase:
         """
         Function to search for requirements from docu lines
 
-        general_pattern -- pattern to search for requirements
+        general_pattern -- pattern to search for requirement comment blocks
         tag -- CB-# tag pattern to be searched in docu
         tag_http -- http pattern to be searched in docu
         tag_http_named -- named http pattern to be searched in docu
         """
-        search_result = general_pattern.search(self.docu_lines)
-        if search_result is None:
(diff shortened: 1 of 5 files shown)


lobster/tools/cpptest/parser/test_case.py
```diff
@@ -168,17 +168,20 @@ def _get_requirements_from_docu_lines(self,
...
-        general_pattern -- pattern to search for requirements
+        general_pattern -- pattern to search for requirement comment blocks
         tag -- CB-# tag pattern to be searched in docu
         tag_http -- http pattern to be searched in docu
         tag_http_named -- named http pattern to be searched in docu
         """
-        search_result = general_pattern.search(self.docu_lines)
-        if search_result is None:
+        blocks = general_pattern.findall(self.docu_lines)
+        if not blocks or blocks is None:
```
SurajBDeore, line 177: findall() always returns a list (possibly empty), so checking blocks is None is unnecessary.

tests-unit/lobster-cpptest/test_cpptest.py
```diff
@@ -13,6 +13,7 @@ class LobsterCpptestTests(unittest.TestCase):
     # unit tests for lobster-cpptest
 
     def setUp(self):
+        self.codebeamer_url = 'https://codebeamer.company.net/cb'
```
SurajBDeore, line 16: We are using this URL in only one test case, so could we please move it to that specific test?

phiwuu (approved):

PR #335, opened 2025-07-02
Refactor `io.py` to simplify unit tests

phiwuu:


lobster/io.py
```diff
@@ -52,7 +53,13 @@ def lobster_write(fd, kind, generator, items):
     fd.write("\n")
 
 
-def lobster_read(mh, filename, level, items, source_info=None):
+def lobster_read(
```
mugdhadhole1, line 56: Similar refactoring can be applied for the `def lobster_write`
Why are we applying it for `def lobster_read` only?

phiwuu: I added the type hints to `lobster_read`, because I came across this function while working on this pull request. I did not come across `lobster_write`, so I did not clean it up.

PR #336, opened 2025-07-04
Introduce `ToolBase` class

phiwuu:
The new class provides an `ArgumentParser` instance and implements the --version flag.

It shall be used as a base class for LOBSTER tools.


tests-system/lobster-cpptest/test_directories.py
```diff
@@ -119,7 +119,7 @@ def test_no_cpptest_file(self):
 
         completed_process = self._test_runner.run_tool_test()
         asserter = Asserter(self, completed_process, self._test_runner)
-        asserter.assertStdErrText('usage: lobster-cpptest [-h] [-v, --version]'
```
kedarnn, line 122: Even though both the flags to check version are added then how is the error message getting modified?

tests-system/lobster-tool-base/asserters.py
```diff
@@ -0,0 +1,54 @@
+from ..asserter import Asserter
+from abc import ABCMeta, abstractmethod
+
+
+EXPECTED_FULL_NAME = "LOBSTER 0.13.1-dev"
```
kedarnn, line 5: Can `FULL_NAME` from `version.py` be used here?

tests-system/lobster-tool-base/asserters.py
kedarnn, line 1: This file can be renamed to `lobster_tool_base_asserter.py`

kedarnn (changes requested):

lobster/tool_base.py
```diff
@@ -0,0 +1,67 @@
...
+            formatter_class=RawTextHelpFormatter,
+        )
+        self._argument_parser.add_argument(
+            "-v",
+            "--version",
+            action="version",
+            default=None,
+            help="print version and exit",
+            version=FULL_NAME,
+        )
```
kedarnn, line 45: Short and Simple 😊

lobster/tools/codebeamer/codebeamer.py
```diff
@@ -469,101 +469,108 @@ def parse_config_data(data: dict) -> Config:
...
+            help=(f"Path to YAML file with arguments, "
+                  f"by default (codebeamer-config.yaml) "
+                  f"supported references: '{', '.join(SupportedConfigKeys.as_set())}'"),
+            default=os.path.join(os.getcwd(), "codebeamer-config.yaml"))
+
+        self._argument_parser.add_argument(
+            "--out",
+            help=("Name of output file"),
+            default="codebeamer.lobster",
+        )
```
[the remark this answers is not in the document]
phiwuu: I have renamed the `ToolBase` class to `MetaDataToolBase`. The next step is to introduce an intermediate class called `ToolBase`, which will add the `--out` and `--config` command line arguments.

I will do this in a separate pull request. Note that some tools use a default value for `--config`, and I am considering to remove this default value. Hence the refactoring would also be a slight change of behavior, so it makes sense to do that in a separate pull request.

kedarnn (approved):

PR #339, opened 2025-07-06
Remove Filter Feature

phiwuu:
Remove the features to use `with <keyword> <string>` in the tracing policy.

Reverts #22 and #33


mugdhadhole1 (commented):
tests-unit\lobster-online-report\data\report-lobster.output
tests-system\lobster-html-report\data\report.output

These two files still contain

      "source": [
        {
          "file": "python.lobster",
          **"filters": []**
        }
      ],

PR #364, opened 2025-07-25
Add lobster-doxygen to the list for C/C++.

BlueAndi:


BlueAndi:
@phiwuu Thanks for adding it. During the weeks the question raised up whether it would make sense something like a short meet&greet to talk about ideas, roadap, methods and etc. What do you think?

PR #379, opened 2025-07-31
Lobster-json system-tests

TannazVhdBMWExt:
Add some new system-tests to lobster-json


tests_system/lobster_json/data/bmw-schema.json
kedarnn, line 1: It would be better not to tell that this is a BMW schema. Please use a different name for the file..

kedarnn (changes requested):

phiwuu (approved):

PR #381, opened 2025-08-01
trlc integration: fix naming

hoe-jo:
fix a naming conflict between package/module **trlc** which leads to not working imports


hoe-jo:
can not be merged, otherwise system tests would break

PR #384, opened 2025-08-04
Fix typos in `README.md` of `lobster-trlc`

phiwuu:


packages/lobster-tool-trlc/README.md
```diff
@@ -57,8 +57,8 @@ example.Codebeamer_Id {
 ```
 
 These functions are applied in order, and we pick the first one that
-fully manages to apply. When a value is `null`, then and it'd be used
-(like in the first to_string function above) then the expansion does
+fully manages to apply. When a value is `null`, and it would be used
+(like in the first `to_string` function above) then the expansion does
 not apply and we move to the next, and so on. When none apply an error
 is created.
```
SurajBDeore, line 63: Suggestion: 
instead of when can we use if?
 When none apply an error is created -->  If none of the functions apply, an error is raised.

kedarnn (approved):

PR #386, opened 2025-08-04
Remove Validation of `custom_data` section

phiwuu:
Removed the function `validate_and_parse_custom_data` for the reasons
given below.
**The overall motivation is to clean the code base and increase the
branch coverage.**

- A JSON validator based on a JSON schema should be used instead of
  implementing a validation logic for each key individually.
- The only benefit so far is that the user recieves a readable message
  if the LOBSTER input file contains data that does not have an
  `items()` function.
  But using `str` as key or value in that dictionary is actually not
  mandatory. Anything that can be converted to `str` works equally fine.
  So the validation is too strict.
- It is okay to remove this feature from the perspective of the tool
  qualification. If the user provides data which is not have iterable
  through an `items()` function, then the tool crashes with a Python
  exception. This is not a convenient user experience, agreed.
  But removing these lines helps to increase the branch coverage,
  and as said above a JSON validator should be used to generate
  helpful messages for the user.
- No unit tests or system tests existed for the function.
- Tools like `lobster-report` and `lobster-online-report` do not
  work with the data provided in `custom_data`. They just store it,
  and serialize it back into their output.
  Only `lobster-html-report` needs data that can be converted to `str`,
  so any validation should happen in that tool.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@
     and generates the lobster report.
     This is similar to running the tool `lobster-report`.
 
+* `lobster-report`
+  - Removed user error messages in case the `custom_data` section contains invalid data.
+    Python will nevertheless exit with return code 1, as previously.
+    The section is not intended to be modified by users.
+    Removing the error messages helps to increase the branch coverage.
+
 ### 0.13.2
 
 * `lobster-html-report`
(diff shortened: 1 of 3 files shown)


SurajBDeore (dismissed):

phiwuu:
I have resolved the merge conflict.

SurajBDeore (approved):

PR #387, opened 2025-08-04
Introduce `Level` data class

phiwuu:


lobster/config/level_definition.py
```diff
@@ -0,0 +1,55 @@
...
+@dataclass
+class LevelDefinition:
+    name: str
+    kind: str
+    traces: List[str] = field(default_factory=list)
+    source: List[Any] = field(default_factory=list)
+    needs_tracing_up: bool = False
+    needs_tracing_down: bool = False
+    breakdown_requirements: List[Any] = field(default_factory=list)
+    raw_trace_requirements: List[Any] = field(default_factory=list)
```
SurajBDeore, line 14: why we don't serialize this data -->  raw_trace_requirements

phiwuu: They are not needed. They are a helper variable.

SurajBDeore (dismissed):

PR #391, opened 2025-08-06
Use API approach in lobster-cpptest

TannazVhdBMWExt:
Add API function

### DIFF
--- a/lobster/tools/cpptest/cpptest.py
+++ b/lobster/tools/cpptest/cpptest.py
@@ -17,13 +17,16 @@
 # License along with this program. If not, see
 # <https://www.gnu.org/licenses/>.
 
+import sys
 from argparse import Namespace
 import os.path
 from copy import copy
 from dataclasses import dataclass, field
-from typing import List
+from typing import List, Union
 from enum import Enum
 import yaml
+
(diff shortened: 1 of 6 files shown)


lobster/tools/cpptest/cpptest.py
```diff
@@ -360,19 +365,46 @@ def __init__(self):
...
-
-        return 0
+        except FileNotFoundError as file_not_found_error:
+            self._argument_parser.error(str(file_not_found_error))
+        except ValueError as value_error:
+            self._argument_parser.error(str(value_error))
+        except KeyError as key_error:
+            self._argument_parser.error(str(key_error))
+        except LOBSTER_Error as lobster_error:
+            self._argument_parser.error(str(lobster_error))
```
kedarnn, line 399: These exceptions are being caught in `_run_impl()` when `_execute()` is triggered then I don't think this is necessary to duplicate them over here.

kedarnn (dismissed):

kedarnn:
Please check if any of the functions that are being called by cpptest uses `sys.exit()`. If any of them exists them please raise an exception instead on it.

mugdhadhole1 (approved):

PR #421, opened 2025-08-20
Robust Handling of Git Submodules in `lobster-online-report`

TannazVhdBMWExt:
Improvements:
- Use the `GitPython` library to obtain information about a git repository.
  The previous implementation called
  `git` as a subprocess, and parsed its output. `GitPython` is widespread
  in use and can be considered reliable.
- Added a `PathToUrlConverter` class which translates a local path into
  a remote git repository URL. It iterates over all submodules known to
  `Repo.submodules` and computes the relative path with respect to that
  submodule. If the path can be computed, then the corresponding submodule
  is found.
  The only public function `path_to_url` returns the full URL to the file
  in the submodule repository.
  The class needs to know the base URL in order to work. This is a
  constructor argument.


kedarnn (dismissed):

kedarnn:
Please add a detailed description for the PR

phiwuu (approved):

PR #423, opened 2025-08-20
Test cleanup

phiwuu:
- Rename class `SpecialAsserter` to `SpecificAsserter`, and  add comments.
- Delete obsolete import statements.
- Include more system tests in `lint-system-tests`

### DIFF
--- a/tests_system/lobster_trlc/test_conversion_errors.py
+++ b/tests_system/lobster_trlc/test_conversion_errors.py
@@ -1,7 +1,6 @@
-from typing import Optional
-from dataclasses import dataclass
-from .lobster_system_test_case_base import LobsterTrlcSystemTestCaseBase
-from ..asserter import Asserter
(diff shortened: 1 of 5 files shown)


tests_system/lobster_trlc/test_conversion_errors.py
```diff
@@ -1,5 +1,3 @@
-from typing import Optional
-from dataclasses import dataclass
 from .lobster_system_test_case_base import LobsterTrlcSystemTestCaseBase
 from ..asserter import Asserter
```
mugdhadhole1, line 2: Now that we have renamed hyphens to underscores
these imports can also be changed to

```suggestion
from tests_system.asserter import Asserter
from tests_system.lobster_trlc.lobster_system_test_case_base import LobsterTrlcSystemTestCaseBase
```

mugdhadhole1 (approved):

PR #445, opened 2025-08-29
add API function to generate HTML report

SurajBDeore:
lobster-html-report: add API function to generate HTML report.

Added a function to allow generating the LOBSTER HTML report via a direct API call, enabling integration with other scripts.

### DIFF
--- a/lobster/tools/core/html_report/html_report.py
+++ b/lobster/tools/core/html_report/html_report.py
@@ -628,5 +628,35 @@ class HtmlReportTool(MetaDataToolBase):
         return 0
 
 
+def lobster_html_report(
+    lobster_report_path: str,
+    output_html_path: str,
+    dot_path: str = None,
+    high_contrast: bool = False,
+    render_md: bool = False
+) -> None:
+    """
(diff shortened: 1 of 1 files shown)


lobster/tools/core/html_report/html_report.py
```diff
@@ -626,6 +626,34 @@ def _run_impl(self, options: argparse.Namespace) -> int:
 
         return 0
 
+def lobster_html_report_api(
+    lobster_report_path: str,
+    output_html_path: str = "lobster_report.html",
```
kedarnn, line 631: Don't give any default value to this and let it be mandatory. We have seen the issues with integration tests where output file is not mentioned but while output comparison specific file name is used which doesn't match the default one.

kedarnn (changes requested):

lobster/tools/core/html_report/html_report.py
SurajBDeore: ok

kedarnn (approved):

PR #461, opened 2025-09-02
udpate requirement in lobster-report

SurajBDeore:


lobster/tools/core/report/requirements/requirements.trlc
```diff
@@ -196,9 +196,10 @@ req.System_Requirement_Aspect Multi_Level_Source_Files {
 
 req.System_Requirement_Aspect Message_Trace_Coverage {
   description = '''
-    IF an item contains a "message" and traces to another item,
-    THEN the traced item shall have 100% coverage,
-    AND the item with the message shall have 0% coverage.
+    IF an item contains a "message",
+    THEN the item shall be counted as "not okay",
+    regardless of whether it traces to another item or not.
+    ELSE the item shall be evaluated according to the general tracing policy.
```
phiwuu, line 202: This needs further investigation.

1. It seems there is a bug, see https://github.com/bmw-software-engineering/lobster/pull/460
2. We need to decide about precedence: If an item has got a "message" AND a "justification", what is the expected item status?

phiwuu (changes requested):

PR #462, opened 2025-09-02
Fixed description for system tests: lobster-cpptest

mugdhadhole1:
- Fixed docstrings where necessary
- Added missing descriptions for test cases


kedarnn (changes requested):
Please keeps docstring first and then below it add the traces.

PR #507, opened 2025-09-17
Support version be traced in the trlc

melodyoncode:


lobster/tools/trlc/converter.py
```diff
@@ -88,9 +88,16 @@ def generate_lobster_object(self, n_obj: ast.Record_Object) -> Optional[Item]:
...
             namespace=rule.lobster_namespace,
             tag=n_obj.fully_qualified_name(),
-            version=None,
         )
 
+        if rule.version:
+            if version := n_obj.field.get(rule.version):
+                item_tag.version = version.to_python_object()
+            else:
+                raise NotImplementedError(
```
phiwuu, line 97: Please use the `item_wrapper` instance. It raises the correct exception when accessing a field that does not exist. The instance is created just a few lines further down. Search for `item_wrapper = ItemWrapper(n_obj)`.

This way you don't need to handle the case if `n_obj.field.get(rule.version)` is not specified.

Also after `version := ...` you are not using `version`, so no need to define that variable.

lobster/tools/trlc/converter.py
```diff
@@ -88,9 +88,16 @@ def generate_lobster_object(self, n_obj: ast.Record_Object) -> Optional[Item]:
         item_tag = Tracing_Tag(
             namespace=rule.lobster_namespace,
             tag=n_obj.fully_qualified_name(),
-            version=None,
```
phiwuu, line 91: Please keep `version=None` over here, as it makes reviewing the code easier. This way it is very explicit that `version` is `None` by default, unless we enter the below if branch.

phiwuu (changes requested):

tests_unit/lobster_trlc/test_converter.py
```diff
@@ -427,3 +427,25 @@ def test_justification_fields(self):
                 lobster_item.just_global,
                 expected_just_global,
             )
+
+    def test_version_field(self):
```
melodyoncode, line 431: add another test (wrong one)

lobster/tools/trlc/requirements/tag_version.trlc
```diff
@@ -0,0 +1,25 @@
...
+      If the conversion rule configuration contains a "version-field: xxx" entry
+      but the record object does not have a field named "xxx",
+      then the tool shall set the tag version to None.
+    '''
+}
+
+req.System_Requirement Tag_Version_None_If_Not_Configured {
+    description = '''
+      If the conversion rule configuration does not contain any "version-field: xxx" entry,
+      then the tool shall set the tag version to None.
```
hoe-jo, line 23: maybe add here:
regardless of the record object

hoe-jo (approved):
LGTM

PR #514, opened 2025-11-17
Use plotly insead of graphviz with adjustment with lobster new version

TannazVhdBMWExt:


tests_system/lobster_html_report/data/custom_data.html
```diff
@@ -34,81 +34,12 @@
   height: 24px;
   vertical-align: middle;
```
SurajBDeore: We are not using this file anywhere.

tests_system/lobster_html_report/test_html_content.py
```diff
@@ -2,9 +2,9 @@
 import unittest
 from tests_system.lobster_html_report.lobster_UI_system_test_case_base import (
     LobsterUISystemTestCaseBase)
-from tests_system.asserter import Asserter
+from tests_system.lobster_html_report.\
```
SurajBDeore: Please use the same approach in all files. In the other file you are importing it. like
from tests_system.lobster_html_report.lobster_UI_system_asserter import (
    LobsterUIAsserter as Asserter)

SurajBDeore:
@TannazVhdBMWExt 
Some checks are not running please check this. I have reviewed this PR once the checks succeed, I will review it again.

tannazVahidiVispiron:
After debugging the reason of hanging the system_tests on windows, we figured it out:

The fact that kaleido can be imported confirms it's installed, but it doesn't prevent the underlying kaleido subprocess from hanging on complex diagrams. This is a known issue with the library.

The multiprocessing code with the timeout is the correct way to handle this

1.Executes the line fig.to_image(format="svg").

2.Launch Subprocess: The plotly library doesn't render the image in your main Python process. Instead, it finds the kaleido executable file that was installed by pip and launches it as a completely separate subprocess. This executable contains its own miniature, built-in Chromium rendering engine.

3.Send Data: Your main Python script serializes your Plotly figure (fig) into a structured JSON format. It then sends this JSON data to the kaleido subprocess through its standard input (stdin).

4.Kaleido Renders: The kaleido subprocess receives the JSON. It uses its internal Chromium engine to render the figure and create the SVG image. This is the critical step where the hang occurs. For complex diagrams, this internal engine can get stuck in a loop or encounter a bug, causing it to stop processing without crashing or sending an error.

5.Python Waits: Back in your main script, the fig.to_image() function is now paused, simply waiting for the kaleido subprocess to send the finished SVG data back through its standard output (stdout).

Why It Hangs Indefinitely
The problem is that your Python process has no insight into what the kaleido subprocess is doing. It's just waiting for data. If the kaleido subprocess gets stuck at Step 4, your Python process will wait forever at Step 5.

This is why a standard try...except block in Python fails to catch the issue—no exception is ever raised in the Python code. The process is just waiting, which is not considered an error.

The multiprocessing solution works because it creates a third process that acts as a "watchdog." It watches the image-generation process and forcefully terminates it if it doesn't finish within the 15-second timeout, preventing your entire application from freezing.

packages/lobster-core/setup.py
```diff
@@ -57,6 +57,7 @@
         "PyYAML>=6.0",
         "yamale>=6.0.0",
         "GitPython>=3.1.30",
+        "plotly==5.10.0",
```
SurajBDeore: do we need to include kaleido?

SurajBDeore:
@TannazVhdBMWExt  
1) GitHub Actions Workflow NOT Updated 
2) Update API documentation to remove dot_path parameter
3) The repository still contains .dot files (Are these still used for documentation purposes, or can they be removed?)

packages/lobster-core/setup.py
phiwuu: yes, I think so

PR #515, opened 2025-11-20
Add lobster-pkg api function

TannazVhdBMWExt:
PR https://github.com/bmw-software-engineering/lobster/pull/513 accidentally deleted the `lobster_pkg` api function from file `lobster/tools/pkg/pkg.py`, because it became a member function of class `PkgTool`.


lobster/tools/pkg/pkg.py
phiwuu: The current proposal does the following:

1. convert all API arguments from native types (like integer) to strings
2. run the lobster-pkt tool, which will parse the strings and convert them to native types (like integer)

In my opinion the lobster-pkg tool should call the API, not the other way around.

phiwuu (changes requested):

SurajBDeore (approved):

kedarnn:
The changes form this PR and suggestion from @phiwuu  is already implemented in https://github.com/bmw-software-engineering/lobster/pull/547.
Hence, this PR is no longer needed and can be closed.

PR #519, opened 2025-12-23
lobster-python: add system test infra and cases

SurajBDeore:
Add runner, base, Bazel targets, tests, and data fixtures for lobster-python system tests, aligning with existing tool infra.

### DIFF
--- a/tests_system/lobster_python/data/basic.py
+++ b/tests_system/lobster_python/data/basic.py
@@ -1,0 +1,22 @@
+import potatolib
+
+
+def trlc_reference(requirement):
+    # lobster-exclude: helper function
+    def decorator(obj):
+        return obj
+    return decorator
+
+
+class Example:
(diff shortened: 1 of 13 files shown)


kedarnn (changes requested):

tests_system/lobster_python/data/basic.py
```diff
@@ -0,0 +1,24 @@
+# --parse-decorator trlc_reference requirement
```
kedarnn: This was not present in the original file https://github.com/bmw-software-engineering/lobster/pull/380/changes#diff-159619bec2900d3e5f1cda2c5fff12186f3c1b1a122d4ce7a7a5d20f88e9454f. 
Why is this added here?

kedarnn (changes requested):

kedarnn (approved):

PR #558, opened 2026-04-07
Migrate LOBSTER Sphinx builder

SebSparrowHawk:


phiwuu (changes requested):
Fork seems to be out of date.
File `manual/tools/pkg.rst` should not be part of this PR.

phiwuu (approved):

PR #571, opened 2026-04-24
lobster-codebeamer improvements

iceman-muc:
add ASIL, ver_ValSetup, ver_ValRationalargumentation extraction and baseline_id support


lobster/common/items.py
```diff
@@ -233,26 +233,45 @@ def __init__(
             name: str,
             text: Optional[str] = None,
             status: Optional[str] = None,
+            asil: Optional[str] = None,
+            ver_ValSetup: Optional[str] = None,
+            ver_ValRationalargumentation: Optional[str] = None,
```
phiwuu, line 238: Unfortunately this feature is against the LOBSTER philosophy. LOBSTER shall remain an agnostic tool that can trace arbitrary artifacts. They don't need to be requirements, they could be architecture elements, Jira tickets or customer bug reports. If attributes like `asil` are hard-coded into `Item`, then we enforce attributes on artifacts which just don't make sense. What is the `ver_ValSetup` of a bug report?

By the way, I assume `ver_ValSetup` means something like "verification and validation setup", but it is not very clear.

If you want to include further information in your LOBSTER item, then please use the `text` attribute. Since https://github.com/bmw-software-engineering/lobster/pull/545 the tool `lobster-html-report` prints this JSON property into the HTML report also for `Activity` objects (not only `Requirement`). You can do so by opening the `*.lobster` file generated by `lobster-report`, enriching the `text` property, and feeding it to `lobster-html-report`.

We are working on a more generic solution regarding the handling of items, see https://github.com/bmw-software-engineering/lobster/pull/524.

Note that you can use markdown inside the `text` property to structure your text, see `--render-md` command line argument: https://github.com/bmw-software-engineering/lobster/blob/bd9a0e4d8ad51340f9e403524b23ab260248c498/lobster/tools/core/html_report/html_report.py#L605

phiwuu (changes requested):

PR #590, opened 2026-06-11
update trlc>=2.0.5

hoe-jo:


packages/lobster-monolithic/setup.py
```diff
@@ -53,7 +53,7 @@
         "miss-hit>=0.9.42",
         "requests>=2.31.0",
         "libcst>=1.1.0",
-        "trlc>=2.0.1",
+        "trlc>=2.0.5",
```
phiwuu: But there is no reason really to enforce 2.0.5
The `lobster-trlc` is fully compatible with `trlc>=2.0.1`

phiwuu (changes requested):

phiwuu (approved):

PR #594, opened 2026-06-23
Replaced os path to pathlib

sachichakrabarti:
Replaced os path to pathlib in 3 files
1. lobster\common\file_tag_generator.py
2. lobster\common\io.py
3. lobster\common\multi_file_input_tool.py


lobster/common/io.py
```diff
@@ -150,6 +150,7 @@ def signal_duplicate_items(
 
 def ensure_output_directory(file_path: str) -> None:
     """Create parent directories for the output file if they don't exist."""
-    output_dir = os.path.dirname(file_path)
-    if output_dir:
-        os.makedirs(output_dir, exist_ok=True)
+    output_dir = Path(file_path).parent
+
+    if output_dir != Path("."):
+        output_dir.mkdir(parents=True, exist_ok=True)
```
phiwuu: We can simplify this even further, because calling `mkdir` for the `.` path is not a problem when `exist_ok=True`.


```suggestion
    Path(file_path).parent.mkdir(parents=True, exist_ok=True)
```

This simplification should increase the branch coverage (if it is covered by tests at all...)

phiwuu (dismissed):

kedarnn (approved):

PR #599, opened 2026-07-09
Catch unexecuted tests in bazel

kedarnn:
Testsuites for each system test tool folder & Script to identify files missing the unittest entrypoint

### DIFF
--- a/tests_system/lobster_codebeamer/BUILD.bazel
+++ b/tests_system/lobster_codebeamer/BUILD.bazel
@@ -40,30 +40,17 @@ py_library(
     ],
 )
 
-py_test(
-    name = "test_extract_requirements",
-    srcs = ["test_extract_requirements.py"],
-    deps = [
-        ":lobster_codebeamer",
-        requirement("flask"),
-    ],
+ALL_TEST_SRCS = glob(["test_*.py"])
+
+test_suite(
+    name = "lobster_codebeamer_tests"
 )
 
-py_test(
-    name = "test_lobster_codebeamer",
-    srcs = ["test_lobster_codebeamer.py"],
-    deps = [
-        ":lobster_codebeamer",
-        "//tests_system",
-        requirement("flask"),
-    ],
-)
-
-py_test(
-    name = "test_valid_flow",
-    srcs = ["test_valid_flow.py"],
+[py_test(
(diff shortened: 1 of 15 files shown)


tests_system/lobster_codebeamer/BUILD.bazel
```diff
@@ -40,30 +40,23 @@ py_library(
...
-    srcs = ["test_valid_flow.py"],
+[py_test(
+    name = src[:-3],
+    srcs = [src],
     deps = [
         ":lobster_codebeamer",
         requirement("flask"),
     ],
-)
+) for src in glob(["test_*.py"])]
```
SurajBDeore:
```suggestion
) for src in ALL_TEST_SRCS
```

tests_system/lobster_codebeamer/BUILD.bazel
```diff
@@ -40,30 +40,23 @@ py_library(
...
-py_test(
-    name = "test_extract_requirements",
-    srcs = ["test_extract_requirements.py"],
-    deps = [
-        ":lobster_codebeamer",
-        requirement("flask"),
-    ],
-)
+ALL_TEST_SRCS = glob(["test_*.py"])
+EXCLUDED_FROM_SUITE = ["test_baseline_id.py"]
```
SurajBDeore: We should add a comment here explaining why we're excluding this. What do you think?

tests_system/lobster_rst_report/BUILD.bazel
```diff
@@ -21,13 +21,16 @@ py_library(
...
+    name = src[:-3],
+    srcs = [src],
     deps = [
         ":lobster_rst_report",
         "//lobster/tools/core/rst_report",
-        "//tests_system",
         "//tests_system/tests_utils",
     ],
-)
+) for src in glob(["test_*.py"])]
```
SurajBDeore, line 36: We have duplicate code in every BUILD file. Can we create a separate macro and call that macro from each tool BUILD file?

kedarnn: We decided to include this change sometime later as it we need to identify the repercussions of it.

SurajBDeore (approved):

PR #606, opened 2026-07-13
Replace trlc make target with bazel target

kedarnn:
- Makefile: Removed the trlc target that used trlc lobster --error-on-warnings --verify
  - CI workflow (.github/workflows/ci.yml):
  - Replaced Python dependency installation with Bazel setup (bazel-contrib/setup-bazel)
  - Updated the trlc invocation to use bazel test instead of make
  - Enabled Bazel caching for improved performance
- Bazel BUILD files:
  - Updated lobster/BUILD.bazel with a new py_test rule for trlc that includes all necessary inputs and dependencies
  - Added trlc_sources filegroups to 11 tool modules to expose TRLC configuration files needed by the bazel test
- .bazelrc: Added test configuration flag to control test outpu

### DIFF
--- a/lobster/BUILD.bazel
+++ b/lobster/BUILD.bazel
@@ -1,2 +1,44 @@
 # BUILD.bazel
 # gazelle:exclude __init__.py
+
+load("@rules_python//python:defs.bzl", "py_test")
+load("//:requirements.bzl", "requirement")
+
+filegroup(
+	name = "trlc_inputs",
+	srcs = glob([
+		"*.rsl",
+		"*.trlc",
+	]) + [
+		"//lobster/tools/codebeamer:trlc_sources",
+		"//lobster/tools/core/ci_report:trlc_sources",
+		"//lobster/tools/core/html_report:trlc_sources",
+		"//lobster/tools/core/online_report:trlc_sources",
+		"//lobster/tools/core/report:trlc_sources",
+		"//lobster/tools/core/rst_report:trlc_sources",
+		"//lobster/tools/cpp:trlc_sources",
+		"//lobster/tools/cpptest:trlc_sources",
+		"//lobster/tools/gtest:trlc_sources",
+		"//lobster/tools/json:trlc_sources",
+		"//lobster/tools/pkg:trlc_sources",
+		"//lobster/tools/python:trlc_sources",
+		"//lobster/tools/trlc:trlc_sources",
+	],
+	visibility = ["//visibility:public"],
+)
+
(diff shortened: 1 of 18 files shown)


lobster/BUILD.bazel
```diff
@@ -1,2 +1,44 @@
...
+		"//lobster/tools/json:trlc_sources",
+		"//lobster/tools/pkg:trlc_sources",
+		"//lobster/tools/python:trlc_sources",
+		"//lobster/tools/trlc:trlc_sources",
+	],
+	visibility = ["//visibility:public"],
+)
+
+py_test(
+	name = "trlc_verify",
```
mugdhadhole1:
```suggestion
	name = "trlc",
```

Can we keep the target name - `trlc`?
Because, we are running the `trlc` tool on whole lobster repository and the make target was also `trlc`.

mugdhadhole1: The Make target showed the output of the trlc tool after executing the command.
eg: `Processed 1 model and 44 requirement files and found no issues`

But this current BUILD target does not show how many files got executed i.e. does not show the `trlc` tool output. I think the behavior should stay equivalent

mugdhadhole1 (changes requested):

lobster/BUILD.bazel
```diff
@@ -1,2 +1,44 @@
...
+		"--error-on-warnings",
+		"--verify",
+	],
+	data = [
+		":trlc_inputs",
+		"@trlc//:cvc5",
+	],
+	deps = ["@trlc//trlc:trlc"],
+	visibility = ["//visibility:public"],
+)
```
hoe-jo, line 44: Should we maybe use the pip package here instead of building it via bazel ourselfes?

lobster/BUILD.bazel
```diff
@@ -1,2 +1,45 @@
...
+	srcs = ["trlc_runner.py"],
+	main = "trlc_runner.py",
+	args = [
+		"lobster",
+		"--error-on-warnings",
+		"--verify",
+	],
+	data = [
+		":trlc_inputs",
+		"@trlc//:cvc5",
```
hoe-jo: I thought we dropped CVC5 in latest TRLC? If required pull in cvc5 via pip?

kedarnn: Yes that's right. No need of this.

mugdhadhole1 (approved):

hoe-jo (approved):

PR #607, opened 2026-07-13
Replace Make targets with BAZEL BUILD targets

mugdhadhole1:
Replaces targets:
- remove-dev
- bump

### DIFF
--- a/util/BUILD.bazel
+++ b/util/BUILD.bazel
@@ -27,6 +27,7 @@ py_binary(
 py_binary(
     name = "bump",
     srcs = ["bump_version_post_release.py"],
+    main = "bump_version_post_release.py",
     visibility = ["//visibility:public"],
     data = ["//:changelog_md"],
     deps = [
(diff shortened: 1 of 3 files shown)


util/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+    data = ["//:changelog_md"],
+    deps = [
+        ":changelog",
+        "//lobster/common:common",
+    ],
+)
+
+py_binary(
+    name = "bump",
+    srcs = ["bump_version_post_release.py"],
```
SurajBDeore, line 29: The target is named bump, so Bazel expects bump.py by default, but the only source is [bump_version_post_release.py]. Please add main = "bump_version_post_release.py"

SurajBDeore, line 29: The bump script updates version.py, but its git add command in bump_version_post_release.py stages lobster/version.py instead. That path does not match the file being modified, so the version change may be left out of the commit. can you also check this

mugdhadhole1: This is fixed in [PR-607](https://github.com/bmw-software-engineering/lobster/pull/612)

SurajBDeore (approved):

PR #613, opened 2026-07-14
ci(bcr): fix BCR publish attestations and sync presubmit template

Rahul-Sutariya:
Update the Publish-to-BCR workflow and BCR presubmit template so future
releases produce a valid, passing BCR entry.

publish.yml:
- Bump bazel-contrib/publish-to-bcr v1.2.0 -> v1.4.1 for the fixed
  attestation-upload path (attaching *.intoto.jsonl to the release and
  immutable-release handling).
- Set download_default_release_artifacts: true so publish pulls the
  archive and attestations from the release_ruleset run and computes
  integrity locally.

.bcr/presubmit.yml:
- Sync the template with the maintained BCR entry: drop windows from the
  main matrix, add a dedicated run_test_module_windows task, and exclude
  environment-dependent tests (Graphviz-based basic:traceability and
  sphinx_rst_report, cpp_focus:fruit_test, lobster_codebeamer, and the
  Windows-only sphinx/html-report/online-report trees).

### DIFF
--- a/.bcr/presubmit.yml
+++ b/.bcr/presubmit.yml
@@ -10,9 +10,8 @@ incompatible_flags:
 
 matrix:
   platform:
-    - ubuntu2204
+    - ubuntu2404
     - macos
-    - windows
   bazel:
     - 8.x
   python_version:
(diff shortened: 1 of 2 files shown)


Rahul-Sutariya:
Hello @bmw-software-engineering/trlc-lobster-maintainers, Cloud you please review this pr?, Thanks!

NikitaJangidBMW (dismissed):

NikitaJangidBMW (approved):

.bcr/presubmit.yml
```diff
@@ -28,12 +27,66 @@ tasks:
...
+      # Windows, so the Sphinx extension tests cannot be analyzed here.
+      - "-@lobster//tests_unit/sphinx_extension/..."
+      # Windows-only: the shared pip hub has no `cffi`, which Selenium (via trio)
+      # requires on Windows, so the html-report tests cannot be analyzed here.
+      - "-@lobster//tests_system/lobster_html_report/..."
+      # Windows-only: the online-report tool converts filesystem paths to GitHub URLs,
+      # and these tests compare against golden files/strings that assume POSIX path
+      # separators ("/") and LF line endings, so they fail on Windows only.
+      - "-@lobster//tests_unit/lobster_online_report/..."
+      - "-@lobster//tests_system/lobster_online_report/..."
```
kedarnn, line 92: Is there no alternative  than disabling the tests?

.bcr/presubmit.yml
```diff
@@ -28,12 +27,66 @@ tasks:
...
+      # images, so this test fails with "please install Graphviz".
+      - "-@lobster//tests_integration/projects/basic:traceability"
+      # projects/sphinx_rst_report:test_sphinx_build runs a full Sphinx build with
+      # warnings-as-errors; the rst_report tool / sphinx.ext.graphviz also need the
+      # Graphviz "dot" binary, which is likewise absent from the BCR CI images.
+      - "-@lobster//tests_integration/projects/sphinx_rst_report:test_sphinx_build"
+      # The lobster_codebeamer system tests stand up a Flask HTTPS mock server on
+      # localhost; they depend on lobster's own provisioned CI environment and do not
+      # run reliably in the sandboxed BCR test environment.
+      - "-@lobster//tests_system/lobster_codebeamer/..."
```
kedarnn, line 55: Is there no alternative than disabling the tests?

kedarnn (changes requested):

.bcr/presubmit.yml
Rahul-Sutariya: They're environmental constraints specific to BCR CI.

The BCR sandbox is minimal and intentionally limited. For example:

- Graphviz is not installed in BCR images, so the Graphviz-dependent tests (basic:traceability and sphinx_rst_report) will always fail there
- The codebeamer tests require a Flask HTTPS server setup that can only run in our full CI environment
- The cpp_focus:fruit_test is intentionally failing—it's used only for metadata generation upstream

All of these tests pass fine in our main CI environment. The exclusions are just for BCR. Installing Graphviz or provisioning infrastructure in the BCR sandbox isn't practical, and there's no way around it without changing how the tests work or the dependencies they need.

.bcr/presubmit.yml
Rahul-Sutariya: These are Windows-specific issues that are hard to avoid without maintaining separate dependency files.

The shared requirements_lock.txt doesn't include Windows-only packages like colorama (needed by Sphinx on Windows) or cffi (needed by Selenium on Windows). Using a universal lock file means these transitive dependencies can get lost.

The online_report and path handling tests also expect POSIX paths and LF line endings, but Windows uses backslashes and CRLF. Those tests pass on Linux and macOS just fine.

Rather than maintain separate lock files for each platform (which adds a lot of complexity), we exclude these specific tests on Windows. They still run in our full CI pipeline where those dependencies are available.

.bcr/presubmit.yml
kedarnn: okay

.bcr/presubmit.yml
kedarnn: okay

kedarnn (approved):

PR #615, opened 2026-07-22
Implement Bazel clean-docs

sachichakrabarti:
1. Replace clean-docs Make target implementation with Bazel target
2. Created util/clean_docs.py

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 SYSTEM_PYTHONPATH:=$(PYTHONPATH)
+BAZEL:=$(or $(shell which bazel),bazel)
 export LOBSTER_ROOT=$(PWD)
 export PYTHONPATH=$(LOBSTER_ROOT)
 export PATH:=$(LOBSTER_ROOT):$(PATH)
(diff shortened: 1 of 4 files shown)


util/clean_docs.py
mugdhadhole1, line 1: Add the license header comment block in this file to be consistent with other files.

Makefile
```diff
@@ -166,7 +167,19 @@ docs:
...
+#touch docs/api_documentation/smoke.txt
+#bazel run //util:clean-docs
+#test ! -d docs
+
+#make equivalent
+#mkdir -p docs/api_documentation
+#touch docs/api_documentation/smoke.txt
+#make clean-docs
+#test ! -d docs
+
```
mugdhadhole1: What does this commented lines do? If they are debug comments then remove these lines

Makefile
```diff
@@ -166,7 +167,19 @@ docs:
 	@sphinx-build -c sphinx -b html . docs/api_documentation
 
 clean-docs:
-	rm -rf docs
```
mugdhadhole1, line 169: Remove this clean-docs target and here after keep only the bazel target.
Replace this make target with bazel target where it is necessary

util/BUILD.bazel
```diff
@@ -41,3 +41,10 @@ py_library(
     srcs = ["changelog.py"],
     visibility = ["//visibility:public"],
 )
+
+py_binary(
```
mugdhadhole1, line 46: Adding a comment about the usage of this target will help.

mugdhadhole1 (changes requested):

Makefile
sachichakrabarti: they are dubug comments. I will remove them.

mugdhadhole1 (approved):

PR #617, opened 2026-07-24
bazel: add Bazel equivalent of make codebeamer-pem

NikitaJangidBMW:
Bazel equivalent of `make codebeamer-pem`. Adds a `genrule` that runs OpenSSL at build time and exposes the generated SSL certificate files as Bazel outputs for the codebeamer mock server.

## Files changed
- BUILD.bazel — `codebeamer_pem` target
- mock_server.py — cert path resolution for Bazel runfiles
- manual-lobster_codebeamer.md — document Make and Bazel cert paths
- .bazelrc — replace deprecated `--experimental_enable_bzlmod` with `--enable_bzlmod`

## How to test
```bash
bazel build //tests_system/lobster_codebeamer/data:codebeamer_pem
bazel test //tests_system/lobster_codebeamer:test_lobster_codebeamer
```

## Note
mock_server.py keeps temporary dual-path support during the migration:
- Bazel: `data/cert.pem` and `data/key.pem`
- Make: `data/ssl/cert.pem` and `data/ssl/key.pem`

The Make fallback can be removed once the remaining system tests are fully Bazel-only.

### DIFF
--- a/tests_system/lobster_codebeamer/mock_server.py
+++ b/tests_system/lobster_codebeamer/mock_server.py
@@ -1,5 +1,6 @@
 import json
 import socket
+from pathlib import Path
 from time import sleep
 from typing import List
 from flask import Flask, Response, request
@@ -12,9 +13,18 @@ import requests
 log = logging.getLogger('werkzeug')
 log.setLevel(logging.ERROR)
 
-# Config
-CERT_PATH = 'tests_system/lobster_codebeamer/data/ssl/cert.pem'
-KEY_PATH = 'tests_system/lobster_codebeamer/data/ssl/key.pem'
+# SSL cert/key lookup differs by test runner:
+# - Bazel runfiles expose cert.pem/key.pem directly under data/
+# - Make writes cert.pem/key.pem under data/ssl/ via make codebeamer-pem
+# TODO: Remove the Make fallback after make system-tests is retired.
+_DATA_DIR = Path(__file__).parent / "data"
+_BAZEL_CERT = _DATA_DIR / "cert.pem"
+_BAZEL_KEY = _DATA_DIR / "key.pem"
+_MAKE_CERT = _DATA_DIR / "ssl" / "cert.pem"
+_MAKE_KEY = _DATA_DIR / "ssl" / "key.pem"
+_USE_BAZEL_LAYOUT = _BAZEL_CERT.exists()
+CERT_PATH = str(_BAZEL_CERT if _USE_BAZEL_LAYOUT else _MAKE_CERT)
+KEY_PATH = str(_BAZEL_KEY if _USE_BAZEL_LAYOUT else _MAKE_KEY)
 MOCK_ROUTE_QUERY_ID = '/api/v3/reports/<int:report_id>/items'
 MOCK_ROUTE_QUERY_STRING = '/api/v3/items/query'
 ARE_YOU_RUNNING_ROUTE = '/are-you-running'
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,2 +1,2 @@
-common --experimental_enable_bzlmod
+common --enable_bzlmod
 test --test_output=all
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,7 @@ integration-tests: packages
 	(cd tests_integration/projects/coverage_zero; make)
 	(cd tests_integration/projects/cpp_focus; make)
 
+# Bazel equivalent: bazel build //tests_system/lobster_codebeamer/data:codebeamer_pem
 codebeamer-pem:
 	@echo "🔐 Generating cert.pem and key.pem for codebeamer system tests..."
 	@mkdir -p tests_system/lobster_codebeamer/data/ssl
--- a/documentation/manual-lobster_codebeamer.md
+++ b/documentation/manual-lobster_codebeamer.md
@@ -74,14 +74,16 @@ Notes:
 
 ## Generating SSL Certificates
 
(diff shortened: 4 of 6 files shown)


lobster.bzl
```diff
@@ -1,11 +1,11 @@
 load("//bazel:providers.bzl", _LobsterProvider = "LobsterProvider")
+load("//bazel/private:generate_certs.bzl", _generate_certs = "generate_certs")
 load("//bazel/private:gtest_report.bzl", _gtest_report = "gtest_report", _subrule_gtest_report = "subrule_gtest_report")
 load("//bazel/private:lobster_gtest.bzl", _lobster_gtest = "lobster_gtest", _subrule_lobster_gtest = "subrule_lobster_gtest")
 load("//bazel/private:lobster_raw.bzl", _lobster_raw = "lobster_raw")
 load("//bazel/private:lobster_test.bzl", _lobster_test = "lobster_test", _subrule_lobster_html_report = "subrule_lobster_html_report", _subrule_lobster_report = "subrule_lobster_report", _subrule_lobster_rst_report = "subrule_lobster_rst_report")
 load("//bazel/private:lobster_trlc.bzl", _lobster_trlc = "lobster_trlc", _subrule_lobster_trlc = "subrule_lobster_trlc")
 
-# Re-export LobsterProvider so it can be loaded from this file
```
sachichakrabarti, line 8: Why was this comment deleted?

bazel/private/generate_certs.bzl
```diff
@@ -0,0 +1,58 @@
+"""Bazel rule to generate self-signed SSL certificates for testing.
+
+This rule generates self-signed X.509 certificates using OpenSSL, suitable for
+HTTPS testing scenarios (e.g., codebeamer mock server).
+
+The generated files (cert.pem, key.pem) are deterministic and cacheable, and
```
sachichakrabarti: Generated files are "deterministic and cacheable" seems incorrect as it generates a fresh RSA key on every invocation (see line 26) hence making output non-deterministic. This also breaks Bazel's action cache.

bazel/private/generate_certs.bzl
```diff
@@ -0,0 +1,58 @@
...
+        ctx: Bazel rule context.
+
+    Returns:
+        DefaultInfo with generated cert.pem and key.pem files.
+    """
+    cert = ctx.actions.declare_file("cert.pem")
+    key = ctx.actions.declare_file("key.pem")
+
+    # Use run_shell to invoke openssl from system PATH
+    # This avoids needing an external openssl Bazel dependency
```
sachichakrabarti: This rule runs OpenSSL from the machine’s PATH. That can work locally but fail in CI or sandboxed/remote Bazel runs. Do you think we should make changes so this target works the same everywhere or that won't be necessary?

lobster.bzl
NikitaJangidBMW: Added the comment back

tests_system/lobster_codebeamer/mock_server.py
```diff
@@ -12,9 +13,19 @@
...
+# Config - SSL certificate paths differ by build system:
+# - Bazel: certs generated by codebeamer_pem target, placed flat in data/
+# - Make: certs committed at data/ssl/
+# TODO: Once make system-tests is fully replaced by
+#       bazel test //tests_system:system_tests,
+#       remove the Make fallback and keep only the Bazel path (data/cert.pem).
+_DATA_DIR = Path(__file__).parent / "data"
+_BAZEL_CERT = _DATA_DIR / "cert.pem"
+_MAKE_CERT = _DATA_DIR / "ssl" / "cert.pem"
+CERT_PATH = str(_BAZEL_CERT if _BAZEL_CERT.exists() else _MAKE_CERT)
```
sachichakrabarti: _BAZEL_CERT.exists() evaluated twice: If cert file appears between two calls (race condition), CERT_PATH would point to Bazel layout but KEY_PATH to Make layout, causing server to fail.
Suggested better fix:
_USE_BAZEL_LAYOUT = _BAZEL_CERT.exists()
CERT_PATH = str(_BAZEL_CERT if _USE_BAZEL_LAYOUT else _MAKE_CERT)
KEY_PATH = str(_DATA_DIR / "key.pem" if _USE_BAZEL_LAYOUT else _DATA_DIR / "ssl" / "key.pem")

lobster.bzl
```diff
@@ -29,3 +30,5 @@ subrule_lobster_html_report = _subrule_lobster_html_report
 subrule_lobster_rst_report = _subrule_lobster_rst_report
 subrule_lobster_report = _subrule_lobster_report
 subrule_gtest_report = _subrule_gtest_report
+
+generate_certs = _generate_certs
```
sachichakrabarti: generate_certs is an internal test utility only needed by the codebeamer mock server. It does not belong here.

lobster.bzl
```diff
@@ -1,4 +1,5 @@
 load("//bazel:providers.bzl", _LobsterProvider = "LobsterProvider")
+load("//bazel/private:generate_certs.bzl", _generate_certs = "generate_certs")
```
sachichakrabarti: see review at line 34, this line is not needed.

bazel/private/generate_certs.bzl
NikitaJangidBMW: Yes, that concern is addressed now. The target no longer runs OpenSSL from PATH; it uses committed fixture cert/key files through Bazel inputs, so it should behave the same in local, CI, and sandboxed/remote runs.

sachichakrabarti (approved):

PR #623, opened 2026-07-30
Add bazel targets under packages folder

sachichakrabarti:
Changed top Makefile (replaced the makefile commands lines 56-93 except clean-packages ) . Removed Makefiles from following sub-folders and added bazel BUILD file + python files. 
Under packages folder :-
1. packages/lobster-core
2. packages/lobster-metapackage
3. packages/lobster-monolithic
4. packages/lobster-tool-codebeamer
5. packages/lobster-tool-cpp
6. packages/lobster-tool-cpptest
7. packages/lobster-tool-gtest
8. packages/lobster-tool-json
9. packages/lobster-tool-python
10. packages/lobster-tool-trlc


mugdhadhole1 (changes requested):
Add license to all the python files

hoe-jo (changes requested):
If I understand the PR correct, you would like to build the wheels with bazel instead of makefile?
Please take a look at this PR https://github.com/bmw-software-engineering/lobster/pull/419 there rules python is used to create the wheel file

PR #633, opened 2026-09-03
Switched to Node 24 on GitHub Actions runners 

phiwuu:
Upgrade GitHub Actions workflows to Node 24-compatible action versions, because Node 20 has been deprecated.
See https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/workflows/REVIEW-INSTRUCTIONS.md
+++ b/.github/workflows/REVIEW-INSTRUCTIONS.md
@@ -1,0 +1,13 @@
+
+# Review instructions
+
+Here are instructions how to review a pull request that changes the commit SHA of a
+Github Workflow Action.
+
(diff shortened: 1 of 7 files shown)


phiwuu:
This PR cannot be merged, because the `clang-tidy` cache is not available. That cache cannot be built, because the built action is still running on Node 20. As a workaround I have created https://github.com/bmw-software-engineering/lobster/pull/635, which will allow us to specify a checkout branch when starting the Workflow in the GitHub GUI.

Update: Now it seems https://github.com/bmw-software-engineering/lobster/pull/633 can be merged and the cache is working. I am a bit confused.

kedarnn (approved):

PR #651, opened 2026-09-11
Rework requirements for `lobster-pkg`

phiwuu:
Formulate potential errors (describing what can go wrong) as system requirements (describing the expected behavior).

Derive system requirements from use cases.
Link system tests to system requirements.
Add missing system tests.

Derive software requirements from system requirements.
Link unit tests to software requirements.
Add missing unit tests.


lobster/tools/pkg/requirements/extraction_rules.trlc
```diff
@@ -1,20 +1,84 @@
...
+        IF the directory of the given output file path does not already exist,
+        THEN the tool SHALL create it, including any missing parent directories,
+        before writing the output file.
+    '''
+    derived_from = [
+        UseCases.List_Requirements_to_Tests,
+        UseCases.List_Requirements_without_Tests,
+        UseCases.List_Tests_to_Requirements,
+        UseCases.List_Tests_without_Requirements,
+        UseCases.Item_Coverage
```
phiwuu, line 71: These use cases don't match for the given requirement. A new use case would make more sense, like
> As a user, I want to specify the output file names (including paths) for any LOBSTER tool so that I can find the output files.

PR #655, opened 2026-09-12
Lift File-Order Restriction in Tracing Policy

phiwuu:
This refactoring is a preparation to support other file formats for the tracing policy for `lobster-report`.

Split `lobster/common/parser.py` Parser into a grammar-only reader that emits a format-agnostic `RawPolicy` (`lobster/common/raw_policy.py`), and a new `lobster/common/policy_builder.py::build_tracing_policy()` that owns all semantic validation (duplicate names, missing source files, self-trace, unknown trace-to/requires targets) and LevelDefinition construction, previously entangled inside
`Parser.parse_level_declaration()/parser.load()`.

**Behavior Change:**
`build_tracing_policy()` uses two-pass validation (collect all level names first, then validate references), lifting the restriction of the file order on 'trace to'/'requires' targets.
`parser.load()` is now a thin wrapper.

No other intended behavior change for currently-valid/currently-rejected configs.


lobster/common/parser.py
Daniele143, line 1: Please update the copyright year in the file header.

lobster/common/policy_builder.py
Daniele143, line 1: I recommend to use type hints

lobster/common/policy_builder.py
```diff
@@ -0,0 +1,88 @@
...
+
+def build_tracing_policy(mh, raw_policy: RawPolicy):
+    levels = collections.OrderedDict()
+
+    # First pass: create every level up front, so later passes can refer to
+    # a level regardless of whether it is declared before or after the
+    # level referencing it.
+    for raw_level in raw_policy.levels:
+        if raw_level.name in levels:
+            mh.error(raw_level.name_loc, "duplicate declaration")
```
Daniele143: Do not use the `MessageHandler` to handle errors. The function shall raise exceptions instead, and the caller shall handle those (e.g. catch them and call `mh.error(...)`).
This makes the function `build_tracing_policy` easier to reuse in other contexts, for example as a LOBSTER API function.

lobster/common/raw_policy.py
```diff
@@ -0,0 +1,61 @@
...
+# lobster/common/policy_builder.py.
+
+from dataclasses import dataclass, field
+from typing import Any, List
+
+
+@dataclass
+class RawSource:
+    file: str
+    loc: Any
```
Daniele143: The `Any` type hint is wrong here.
The actual type is `Location` from `lobster.common.location`.

Daniele143 (changes requested):
Increase the LOBSTER version to 1.1.0, as lifting the file order constraint is a real feature improvement.

lobster/common/policy_builder.py
phiwuu: Fixed

lobster/common/raw_policy.py
phiwuu: Fixed

lobster/common/parser.py
phiwuu: updated

Daniele143 (approved):

PR #665, opened 2026-09-17
Rework requirements for `lobster-html-report`

phiwuu:
Formulate potential errors (describing what can go wrong) as system requirements (describing the expected behavior).

Derive system requirements from use cases.


lobster/tools/core/html_report/requirements/html_report_content.trlc
```diff
@@ -1,35 +1,46 @@
 package html_req
 import req
+import UseCases
+
+req.System_Requirement HTML_Report_Uses_Valid_HTML_Syntax {
+    description = '''
+      The tool shall generate an HTML report that adheres to valid HTML syntax.
+    '''
+    derived_from = [UseCases.HTML_Output]
```
PratikshaJZade, line 9: Use Case is irrelevant for the above system requirement. Need to create a new use case for html syntax.

lobster/tools/core/html_report/requirements/html_report_content.trlc
```diff
@@ -1,35 +1,46 @@
...
 
 req.System_Requirement Meta_Tag_In_Html_Report {
     description = '''
-      IF a HTML report is generated,
-      THEN the file should contain meta tag,
-      HAVING attributes "http-equiv" and "content".
+      The HTML output file shall contain the meta tag
+      with attributes "http-equiv" and "content".
     '''
+    derived_from = [UseCases.HTML_Output]
```
PratikshaJZade, line 17: Use Case doesn't mention about tags, attributes and so is irrelevant to above requirement.

lobster/tools/core/html_report/requirements/html_report_content.trlc
```diff
@@ -1,35 +1,46 @@
...
 
 req.System_Requirement Item_Data_Unique {
     description = '''
-      IF the HTML report tool processes multiple items with their respective data,
-      THEN the tool shall ensure that data is not mixed between items,
-      AND each item shall be processed correctly with its unique data.
+      The tool shall print all items from the input file into the HTML output file,
+      including their carried data.
     '''
+    derived_from = [UseCases.HTML_Output]
```
PratikshaJZade, line 25: Use Case is irrelevant, need use case to map with above requirement.

PratikshaJZade (changes requested):

lobster/tools/core/html_report/requirements/html_report_content.trlc
phiwuu: The intention here was to have a requirement to enforce HTML according to some standard, like HTML5 or HTML 4.01 Strict. The use case just requests "HTML output", and the system requirement here enforces valid HTML. But the use case does not say whether it should be HTML 4 or 5 (and that is irrelevant to the use case), but the tool must make a decision on the HTML standard. So we need a system requirement that enforces valid HTML.

lobster/tools/core/html_report/requirements/html_report_content.trlc
phiwuu: same argumentation as above

lobster/tools/core/html_report/requirements/html_report_content.trlc
phiwuu: The use case specifies that a traceability report shall be created as HTML. That means, all the information of "a traceability report" must be present in the HTML file. So the use case is correct in my opinion.

lobster/tools/core/html_report/input_file.trlc
```diff
@@ -1,12 +1,14 @@
...
 import req
+import UseCases
 
 req.System_Requirement Missing_Lobster_File {
     description = '''
       IF the input .lobster file does not exist or is missing,
       THEN the tool shall exit with a non-zero return code (2)
       AND shall print an error message to STDERR indicating that the specified file is not a valid file.
     '''
+    derived_from = [UseCases.HTML_Output]
```
PratikshaJZade: Use Case is irrelevant here. We can trace above Requirement to newly created use case for Invalid input.

